PR #3, opened 2025-01-31
Quadtree and Quadtree Viewer implementation

IDarkgenesis:
Added Quadtree and Quadtree Viewer. When running the engine a new viewport appears with a quadtree rendered, it can be closed and reopened under General tab in the menu.

How to test:

Check functions and change parameters inside QuadtreeViewer inside the UI filter, there is the creation of the quadtree and the insertion of points, as well as the draw call.

### DIFF
--- a/Game/imgui.ini
+++ b/Game/imgui.ini
@@ -32,22 +32,22 @@ Collapsed=0
 
 [Window][Sence]
 Pos=0,19
-Size=1076,701
+Size=1014,701
 Collapsed=0
 DockId=0x00000004,0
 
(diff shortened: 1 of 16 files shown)


SirKaza:
Rename:
· class QaudtreeViewer to QuadtreeViewer 
· in Application.h method GetDebugDreawModule to GetDebugDrawModule

Quadtree viewer doesn't resize, don't know if this is expected.

Apart from these things, everything seems clear to me. I tested different values and it works as expected. For now without a scene it's more than enough.

Great work!

PR #6, opened 2025-02-05
Added Optick profiler library

IDarkgenesis:
Added Optick profiler library.

To test:

Download last version of [Optick](https://github.com/bombomby/optick/releases), compile and run the engine. While the engine is running execute the Optick.exe and click the play button at the top left corner and stop to check everything is working.


SirKaza (approved):
Great work, I tested it, and it works perfectly fine as teachers showed us in class. For now, reviewing only application is good enough. About the frame buffer fix well done, now we x2 or even more the frames/frame time.

PR #9, opened 2025-02-08
Import dialog

SirKaza:
New option added in main menu window "Import", opens a new window that is basically a simple file explorer. 
This dialog has the following features:
· Navigate through paths.
· Go to drives folder.
· Search introducing text for a specific file.
· Select file to import.

Import functionality has to be connected with binaryFiles branch, this is only the UI.


srLlobet:
look good

srLlobet (approved):

PR #14, opened 2025-02-14
Add EngineTimer and GameTimer

PolCurto:
EngineTimer is the same as the timer we had now. I have placed it in Application but could be in main. It is unstoppable and gives the deltaTime to all the update functions.

GameTimer can be modified with timeScale, paused, stopped and played again. The step only advances one frame and can only be done when the timer is paused. The Tick() and Step() functions return the deltaTime, which is also affected by the timeScale. The GameTimer also has a realTime timer, which simply counts the real time the game is executing. This timer is accessible from other classes with Application->GetGameTimer(). I guess it should give the deltaTime to the gameObjects when the game executes (or something like that).

There is a window in the EditorSettings UI to see how the time is being counted.


LukasGanter (changes requested):
- Destructor in EngineTimer needs to be virtual so the one in GameTimer gets called
- UpdateTimes in GameTimer can be const and possible inlined
- EngineTimer class is not in a folder
- Reference time is reset on stop but reinstated when clicking play again. On purpose?
- Time scale half -> Should the delta time not be doubled then?
- Bug: Play pressed -> Paused -> Time scale to 2 -> Play -> Game time negative

SirKaza:
If play is pressed while in pause status, game time resets, is this expected?

LukasGanter (approved):

PR #16, opened 2025-02-14
Limit Framerate

SirKaza:
Features implemented:
· Added some information about the engine
· Vsync checkbox
· Slider for fps

Now in editor settings window from menu there is an Application section which has some information about the engine, a slider for selecting the frame rate and frames graph display. 
In window section there is a checkbox for enable/disable vsync.
If vsync disabled and slider for fps = 0 → unlimited fps.


IDarkgenesis:
Tested the vsync and unlimited FPS and works properly. Also good change of some functions to be inline.

Good job!

IDarkgenesis (approved):

PR #17, opened 2025-02-15
Editor input, hardware & aboutUs

SirKaza:
· Added input keyboard, mouse buttons and mouse information in runtime.
· Added hardware specs information, as we had in the first assignment.
· Added new menu option About. With some description about our engine, reference links, libraries and build/config options.
· Also fixed a warning from limit fps and added const.

How to test:
About window is in the top menu bar.
Hardware and input are collapsed widgets inside editor settings.


IDarkgenesis (approved):
Tested the features, everything worked as expected

PR #21, opened 2025-02-17
Normal map works

PolCurto:
Changed a variable in materialGPU to know in the shader whether the material has normals or not.

Now the lights Render() is done in LightsConfig, so frustum culling is avoided.

Currently tangents are not generated by the engine, so meshes with no tangents are not affected by light. This can be checked with the delivery scene: the objects illuminated have a tangent primitive in the .gltf, while the others don't. In theory the meshes the artists export will have tangents generated by blender, but it would be nice to have a way to generate them in the engine just in case.


IDarkgenesis (approved):
Tested and haven't seen any problems

LukasGanter (approved):

PR #23, opened 2025-02-21
Lights loading/saving THO-66

caspi99:
You can edit these parameters and save the scene. It will save it and load from the json.
image


SirKaza (changes requested):
LibraryModule when savingScene check if lightsConfig is nullptr, if it is dont continue with lightsConfig info (line 111).

Also, merge main branch into this branch and resolve conflicts, make sure everythings works after merge.

SirKaza (approved):
Nice job, feels better now with encapsulation.

PolCurto (approved):
Lights config is correctly saved and loaded. Everything else seems to be working fine, good job

PR #24, opened 2025-02-21
Fixed Drag & Drop Scene GameObject

MarcLopezS:
Has been disabled drag and drop when the source gameObjects is the Scene GameObject (Root GameObject).


SirKaza:
Well done, I suppose that drag n drop to another game Objects and conserving his local and not moving should be another task, but I can't find it.

SirKaza (approved):

MarcLopezS:
> Well done, I suppose that drag n drop to another game Objects and conserving his local and not moving should be another task, but I can't find it.

Yes, that will be for another task that should be planned to do. I can't find it either. I think is still missing to create.

PolCurto (approved):
Works as expected. By the way, dragging a gameObject to one of its children crashes the game, but I guess that will be another task

PR #25, opened 2025-02-23
Bug/direction property duplicate THO-178

PolCurto:
Removed the direction property from both DirectionalLight and SpotLight. When calling GetDirection(), it returns a default direction vector (0,-1,0) rotated by the globalTransform rotation. To test it, you can confirm if rotating the gameObject properly rotates the lights, both with gizmos enabled or disabled.


SirKaza (changes requested):
I tested with another branch, and now it works as expected, rotation without gizmos and not breaking with -INF values. Nice work!

Although, there are 2 warnings building solution regarding the method GetDirection() in direction and spotlight.

SirKaza (approved):

IDarkgenesis (approved):
Tested the changes and works as expected.

PR #26, opened 2025-02-23
Bug/rename certain components THO-176

PolCurto:
Renamed DirectionalLight, PointLight and SpotLight. I have not seen more components wrongly named, but you can check if there are others just in case. Also, you can test if everything works properly, but with the changes I've done if there were some errors it would simply crash


SirKaza (approved):
Good changes!

IDarkgenesis (approved):
Haven't seen any issues with the changes.

PR #27, opened 2025-02-24
bug/loadTwice THO-177

SirKaza:
Fixed loading twice a mesh or material crashed the application. 
Also, I was debugging and thinking that if we close the scene, all resources should be deleted.

### DIFF
--- a/SobrassadaEngine/Scene/Components/Standalone/MeshComponent.cpp
+++ b/SobrassadaEngine/Scene/Components/Standalone/MeshComponent.cpp
@@ -105,14 +105,15 @@ void MeshComponent::AddMesh(UID resource, bool reloadAABB)
 {
     if (resource == CONSTANT_EMPTY_UID) return;
 
+    if (currentMesh != nullptr)
+    {
+        if (currentMesh->GetUID() == resource) return;
+    }
+
     ResourceMesh* newMesh = dynamic_cast<ResourceMesh*>(App->GetResourcesModule()->RequestResource(resource));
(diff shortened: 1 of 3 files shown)


PolCurto (approved):
Everything seems to be working fine now, good job

SobrassadaEngine/Scene/Components/Standalone/MeshComponent.cpp
```diff
@@ -105,6 +105,11 @@ void MeshComponent::AddMesh(UID resource, bool reloadAABB)
 {
     if (resource == CONSTANT_EMPTY_UID) return;
 
+    if (currentMesh != nullptr)
+    {
+        if (currentMesh->GetUID() == resource) return;
```
IDarkgenesis, line 110: This line can be merged with the if on top for just a single if statement

SobrassadaEngine/Scene/Components/Standalone/MeshComponent.cpp
```diff
@@ -128,6 +133,11 @@ void MeshComponent::AddMesh(UID resource, bool reloadAABB)
 
 void MeshComponent::AddMaterial(UID resource)
 {
+    if (currentMaterial != nullptr)
+    {
+        if (currentMaterial->GetUID() == resource) return;
```
IDarkgenesis, line 138: This line can be merged with the if on top for just a single if statement

IDarkgenesis (changes requested):

PR #28, opened 2025-02-25
Basic engine editor window tho 131

DHazardFenrir:
Added the basic window in a class to show an empty window and close.


SobrassadaEngine/SobrassadaEngine.vcxproj
```diff
@@ -137,6 +137,81 @@
   </ItemDefinitionGroup>
```
IDarkgenesis, line 137: Please have a look at VS config, this modifications won't allow to compile the Engine in release mode

IDarkgenesis (changes requested):
The creation of one window works, but if two are created and assigned to the same anchor point the engine crashes

WindowEditorCrash

SobrassadaEngine/SobrassadaEngine.vcxproj
LukasGanter: @IDarkgenesis How did you solve that issue last time?

IDarkgenesis: Manually removing the configs inside the vcxproj, but this should not appear in the first place

LukasGanter: As the following tasks for this issue are very important, my suggestion would be to merge this now even though it might not build and fix the build in a bug task in the next sprint.

LukasGanter (approved):

PR #30, opened 2025-02-26
Load gltf defaults THO-129 

srLlobet:
Grabbing default transforms from gltf files and applying them to gameObjects.


srLlobet:
To test: delete current asset and library files (except skybox) and re-drag gltf files to the engine. The generated library files now contain the gltf's transform and apply it to the mesh.

SobrassadaEngine/Scene/Components/Standalone/MeshComponent.cpp
```diff
@@ -91,12 +91,22 @@ void MeshComponent::Render()
...
 
-        float4x4 model         = float4x4::FromTRS(
+        float4x4 localMatrix   = float4x4::identity;
+        localMatrix            = float4x4::FromTRS(
+            localTransform.position,
+            Quat::FromEulerXYZ(localTransform.rotation.x, localTransform.rotation.y, localTransform.rotation.z),
+            localTransform.scale
+        );
+        /*
+        float4x4 globalMatrix = float4x4::FromTRS(
```
IDarkgenesis, line 101: This is the correct matrix to use, the global transform, right now if you move the root component using gizmos the mesh stays in the same place, local transform should be ignored

IDarkgenesis (changes requested):

PR #32, opened 2025-02-27
Globals unification tho 175

jmartiolivart:
After a thorough review of the code, I confirmed that all constants (including const, constexpr, and #define) are correctly scoped. External libraries such as debug_draw.hpp and DirectXTex have their own constants, which remain untouched and are not candidates for inclusion in Globals.h. 
Only global constants that aren’t local should be moved, and all local constants are appropriately encapsulated within their respective modules (e.g., Camera, FrustumPlanes, Octree). I also examined the usage of magic numbers and #define directives, and everything is correctly placed.
As a result, only RAD_DEGREE_CONV constant has been changed for existing DEGTORAD (with addient conversion)


IDarkgenesis (approved):
Nothing breaks with this changes

SirKaza (approved):
Don't see any problems, good job reviewing all global variables.

PR #33, opened 2025-02-28
Support all gltf files THO-128

jmartiolivart:
The engine now supports both material workflows. It chooses the appropriate workflow based on the glTF file content, ensuring that materials using either KHR_materials_pbrSpecularGlossiness or pbrMetallicRoughness are imported correctly.

Conducted a detailed review of the glTF import code and confirmed that the current implementation already assigns default values when certain attributes are missing. For example, if the "NORMAL" attribute is absent, the code assigns a default value of (0,0,0).

As upcoming features (e.g., animation support) are implemented, this part of the import process will be revisited to ensure it meets any new requirements.


IDarkgenesis (commented):
Plese add a comment on how to test the changes, have tried to load the baker house in a different branch and the texture does not render properly, in this branch it does, if that is what should happen, then changes work properly.

jmartiolivart:
Yes, the Baker House GLTF is an older format. With these changes, the engine now properly supports that format. If everything is working correctly, you should see the meshes rendered with the correct material and texture using the pbrMetallicRoughness workflow.

For testing, please try loading models with both the old and new GLTF formats. You can use some sample assets from KhronosGroup's glTF-Sample-Assets repository:
https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models

This should help verify that the textures and materials render correctly across different asset types.

IDarkgenesis (approved):
Works perfectly, is also able to load a model from the artists good job!

caspi99:
When loading from an scene, I have this error when getting a normal Texture
LibraryModule.cpp(364) : obtained path: Library/Textures\1119015567097783.ddsAssertion failed!

Program: SobrassadaEngine.exe
Resourc.eTexture.cpp
Line: 42

Expression: false && "Unsupported format"

caspi99 (changes requested):
I think when changing the material.h, the .mat brokes

jmartiolivart:
Reordered for not breaking the .mat existing (when loading).

caspi99 (approved):
Now its working good, good job!

PR #34, opened 2025-03-01
Bug/input editor THO-169

SirKaza:
Things done:
· Camera controls only in scene window.
· ImGuizmo manipulation only in scene window.
· ImGuizmo negative values fixed.
· Focus scene window with right click.

How to test: 
Try inputs outside scene window (focus "f", camera "o", gizmos "w" "e" "r") or moving objects outside. If the window is not focused (left or right click) and hovered with the mouse, you won't be able to do scene controls. Also, test writing and check if it's not doing inputs for scene. ImGuizmo negative values becoming crazy fixed.


PolCurto (approved):
Everything seems to work as it should

PR #35, opened 2025-03-02
Gizmos global mode THO-36

LukasGanter:
Testing needs to be done with multiple components in a hierarchy. All components can be added.
Keys E (Local) and R (World) can be used to switch between the world and local coordinate systems.
Keys G (Position), H (Rotation) and J (Scale) are used to switch the gizmo between the different modes.
All inputs can be made via the input fields or the gizmo.
[Class diagram](https://miro.com/app/board/uXjVLsKWVpo=/?moveToWidget=3458764619268174967&cot=14)

### DIFF
--- a/SobrassadaEngine/Scene/Components/Component.cpp
+++ b/SobrassadaEngine/Scene/Components/Component.cpp
@@ -143,7 +143,8 @@ bool Component::DeleteChildComponent(const UID componentUID)
 
 void Component::RenderGuizmo()
 {
-    if (App->GetEditorUIModule()->RenderImGuizmo(globalTransform))
+    const Transform& parentTransform = GetParentGlobalTransform();
+    if (App->GetEditorUIModule()->RenderImGuizmo(localTransform, globalTransform, parentTransform))
     {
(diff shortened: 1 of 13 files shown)


IDarkgenesis (changes requested):
Works as expected but, please try find another keybind combination for changing between transforms, E is already used for lifting the camera up.

Also, it feels weird that the local transform is updated when the global transform is modified, shouldn't it keep its transform since its local? Right now there is no difference between both, they behave the same way.

LukasGanter:
> Works as expected but, please try find another keybind combination for changing between transforms, E is already used for lifting the camera up.
> 
Input binding changend to R (Local), T(World)

> Also, it feels weird that the local transform is updated when the global transform is modified, shouldn't it keep its transform since its local? Right now there is no difference between both, they behave the same way.

Only the local transforms are stored, so it is necessary to change the local if the world is manipulated. 
Also they are different, but only if one of the parents has a non-default Transform.
The local transform is always calculated as the difference between the global transform and the parent transform.

IDarkgenesis (approved):

PR #36, opened 2025-03-03
Asset meta files THO-48

srLlobet:
To test: delete the whole library and reimport all files. Meta files are readable so just open them and check. They should contain the UID, the path to the original resource, the last modified time and importer-specific settings.

Has been testes by me by adding models, checking that they work correctly, then closing and opening and, without reimporting the gltfs, the assets are still accessible, now by name instead of UID


srLlobet:
Materials now use primitive name, and the filters are fixed

SirKaza (changes requested):
Filters in Visual Studio are missing.

SirKaza (changes requested):
Change filenames for files and meta to UID, and in meta files save the name.

SirKaza:
Still making changes, don't review this PR yet.

SirKaza:
Greenlight for reviews

SirKaza:
Things done:
Assets files like meshes, textures and materials now create a metadata file (.meta) with UID, name, assetPath, lastModifiedTime and importedSettings. Now library its not important and can be deleted.
LastModifiedTime is being saved properly but is not being used (future task).

How to test:
In assets folder there is two folders: scene and metadata.
Try importing things, creating an scene and saving it. Then you can delete library folder.
Then run engine and library files will be reimported and scene should display as expected.

IDarkgenesis (approved):
It works good job! 🧇

SirKaza (approved):
Seems good 👍

PR #38, opened 2025-03-04
Components in gameObjects THO-34

MarcLopezS:
Added recodification, in order for the gameObjects now have the responsibility to manage their own GameObjects. Each gameObject has a map of their own components.


MarcLopezS:
There is a bug reported when a new scene is created, Hierarchy and component window is not shown. This bug is skipped because will be solved in another branch

PR #39, opened 2025-03-05
Lg gizmos global mode tho 36

LukasGanter:
Testing in the same way as described [here](https://fpcvideogames.youtrack.cloud/issue/THO-36/Gizmos-Global-mode#focus=Comments-5-29.0-0)
Two known issues: 
- If the scale input is locked and the user wants to input i.e. 0.5, then the first digit 0 will lead to all scale axis being 0, basically losing its scale. Can be avoided for now by typing .5 instead of 0.5
- Meshes can become sheared if many different translations, rotations and scales are applied to a parent and child
I suggest I fix these issues in a bug task once this one is submitted, as other tasks are blocked by this task and the issues don´t affect basic functionality.

I will solve the merge conflicts while testing already start and also make the class diagram once the merge conflicts are solved.


PolCurto (changes requested):
It looks good, the transforms behave as expected. However it should be merged with the main branch first

PolCurto (approved):
Works properly

PR #41, opened 2025-03-05
Library integration recast and detour THO-193

srLlobet:
Added Recast and Detour. There are also extra folders with more functionalities, but we can add them to the project later if we need them.


SirKaza (changes requested):
Includes are only in debug. Should be also in release.
Change pull request name from tho 193 to THO-193 so its tracked in YouTrack.

SirKaza (changes requested):
Release build errors with recast includes.

SirKaza:
Merge Conflicts

caspi99 (changes requested):
You should add the libraries version in the "About Sobrassada" and the license in the Licenses folder image. The rest is fine

PolCurto (approved):
Nice

caspi99 (approved):
Now its good. Good job!

PR #42, opened 2025-03-06
Resource skeleton THO-142

PolCurto:
What to test: When importing a .gltf, a ModelResource will be created. It is stored in JSON format in the models directory to check if it is correct. When in a scene, in a new window in the top bar models can be loaded. It should load everything the .gltf had inside, creating a gameObject per node and assigning the corresponding meshes and materials if they need. 
Right-clicking a gameObject will show an option to toggle a debug view: this is mostly temporary to debug if the bones are placed in the correct position, with the skinning and animation editor it is probable that it will be removed.


SobrassadaEngine/Modules/SceneModule.cpp
```diff
@@ -7,6 +7,7 @@
 #include "GameObject.h"
 #include "LibraryModule.h"
 #include "Octree.h"
+#include "Model.h"
```
SirKaza, line 10: it is needed?

SobrassadaEngine/Scene/GameObjects/GameObject.cpp
```diff
@@ -5,6 +5,8 @@
 #include "Root/RootComponent.h"
 #include "SceneModule.h"
 
+#include "DebugDrawModule.h"
```
SirKaza, line 8: This include with top group

SobrassadaEngine/Scene/GameObjects/GameObject.cpp
```diff
@@ -48,7 +50,7 @@ GameObject::GameObject(const rapidjson::Value& initialState) : uuid(initialState
 GameObject::~GameObject()
 {
     App->GetSceneModule()->RemoveComponent(rootComponent->GetUID());
-    delete rootComponent;
+    if (rootComponent != nullptr) delete rootComponent;
```
SirKaza, line 53: If not needed, c++ manages that

SobrassadaEngine/Utils/Scene.cpp
```diff
@@ -11,6 +11,13 @@
 #include "OpenGLModule.h"
 #include "SceneModule.h"
 
+#include "ResourcesModule.h"
```
SirKaza, line 14: This group with top group, no separation needed

SobrassadaEngine/Utils/Scene.cpp
```diff
@@ -93,6 +100,11 @@ update_status Scene::Render(float deltaTime)
         }
     }
 
+    for (const auto& gameObject : gameObjectsContainer)
+    {
+        gameObject.second->DrawGizmos();
```
SirKaza, line 105: Why gameObjectsContainer and not objectsToRender?

SobrassadaEngine/Utils/Scene.cpp
```diff
@@ -358,3 +370,58 @@ Component* Scene::GetComponentByUID(uint64_t componentUID)
...
+
+            if (nodes[i].meshes.size() > 0)
+            {
+                GLOG("Node %s has %d meshes", nodes[i].name.c_str(), nodes[i].meshes.size());
+
+                for (const auto& mesh : nodes[i].meshes)
+                {
+                    MeshComponent* meshComponent =
+                        reinterpret_cast<MeshComponent*>(ComponentUtils::CreateEmptyComponent(
+                            COMPONENT_MESH, LCG().IntFast(), gameObject->GetRootComponent()->GetUID(),
```
SirKaza, line 406: Instead of LCG().IntFast() use GenerateUID()

SirKaza (changes requested):
Tested with ZomBunny and displays as expected, some of the bones are not perfect, but I suppose is normal.
Please take a look at my feedback via comments.
Also, fix all the build warnings in GetNodeTransform() method.

SobrassadaEngine/Utils/Scene.cpp
PolCurto: Because only objects with a mesh are in objectsToRender, and bones have no mesh

SirKaza: True, I remember you told me that, thanks

SirKaza (approved):
Nice, the way models are being set in scene with Lukas (THO-200) changes will not be possible (2 meshes in same game object), so changes will be needed soon about this.

PolCurto:
Changed the loading of models so they work with the new scene management

caspi99 (approved):
Good Job! ♿ ♿ ♿ ♿

PR #48, opened 2025-03-08
Camera component THO-41

caspi99:
For testing:
You can add a camera component. In the ImGui you can edit the camera type, the fov and the near/far plane and you can activate/deactivate the guizmos and say what camera is the main one for the game. You can move the camera and rotate it. You can save the camera in a scene and load from one and you can click on Play and see that it changes to the main camera you created.
Dont try to move the camera on Play because the component doesn't have to have any kind of movement, its needs to be controlled by the player. If you press stop it will crash as the editor mode switch is in testing for youtrack.


LukasGanter:
Please resolve all the conflicts with main

LukasGanter:
- For the orthographic view the view area is vertical, not horizontal, Also the debug cube is rendered with the camera at the center
- Loading the scene after saving doesn´t show the camera view cone for me anymore. That might be in relation to a bug with the octree though which is fixed in another branch
- The components local transform is used for transforming, it is necessary to use the global transform to also incorporate transforms for the parents
- Removal of camera component crashes the engine (Might be fixed in THO-200)
- Default near / far for orthographic are the same, which leads to confusion as nothing is being rendered when selecting that option
Example of orthographic

I would suggest you merge lg_RemovalComponentHierarchy_THO-200 into your branch and adapt your code to it. That might fix most of the issues i found.

LukasGanter (changes requested):
- No debug view is rendered for me
- Loading the scene after saving doesn´t show the camera view cone for me anymore. That might be in relation to a bug with the octree though which is fixed in another branch
- The components local transform is used for transforming, it is necessary to use the global transform to also incorporate transforms for the parents
- Removal of camera component crashes the engine (Might be fixed in THO-200)
- Default near / far for orthographic are the same, which leads to confusion as nothing is being rendered when selecting that option
Example of orthographic
Example of what happens with a rotated orthographic camera

I would suggest you merge lg_RemovalComponentHierarchy_THO-200 into your branch and adapt your code to it. That might fix most of the issues i found.

PR #49, opened 2025-03-09
Lg removal component hierarchy THO-200

LukasGanter:
Removed the component hierarchy and replaced it with a simple list of components. 
Every component type can only added once to each gameObject, tries to add an already existing one are ignored.
Scene saving and loading was also modified. 
Testing needs to include adding / deleting gameObjects / components, saving and loading scenes as well as if the transforms and bounding boxed are calculated correctly. 
Because of the many changes i will adapt the class diagram once this pr has been approved.

Known issue: Reordering gameObjects in the same branch will crash the engine. A fix for that is already developed in a separate bug task and will be applied once this task has been merged.


SobrassadaEngine/Utils/Scene.cpp
```diff
@@ -291,24 +365,17 @@ void Scene::RemoveGameObjectHierarchy(UID gameObjectUUID)
     delete gameObject;
 }
 
-void Scene::RemoveComponent(uint64_t componentUID)
+const std::unordered_map<uint64_t, Component*> Scene::GetAllComponents() const
```
SirKaza, line 368: UID

SobrassadaEngine/Utils/Scene.cpp
```diff
@@ -291,24 +365,17 @@ void Scene::RemoveGameObjectHierarchy(UID gameObjectUUID)
...
-
-AABBUpdatable* Scene::GetTargetForAABBUpdate(UID uuid)
-{
-    if (gameObjectsContainer.count(uuid))
-    {
-        return gameObjectsContainer[uuid];
-    }
-
-    if (gameComponents.count(uuid))
+    std::unordered_map<uint64_t, Component*> collectedComponents;
```
SirKaza, line 370: UID

SirKaza (approved):
Looks amazing, thanks for your effort and changing it quickly.
Take a look at the two comments right below before merging.

IDarkgenesis (changes requested):
Quitting the engine with the menu or closing the app produces a crash, only if there are meshes loaded.

image

LukasGanter:
> Quitting the engine with the menu or closing the app produces a crash, only if there are meshes loaded.
> 
> image

Fixed it.
This is actually a important thing to consider. The destructors are called in the wrong order, we need to use the shutdown function for all cleanup

IDarkgenesis (approved):
Good job! Nice fix! 🥇

PR #50, opened 2025-03-09
task/Render Information THO-62

SirKaza:
Things done:
· Added render information section in EditorSettings/OpenGL window. This section contains amount of draw calls, triangles per second and vertex visibles.
· Inline OpenGL setters.
· Also, changed metadata primitives names for meshes (instead of third primitive: mesh012 now is: mesh2).
· Metadata extension now until is .smeta.

How to test:
Try loading a scene and inserting meshes. Then open in menu View -> EditorSettings -> OpenGL and watch render information. If camera is moved information should change if object is not rendered for example.


SobrassadaEngine/ResourceManagement/Resources/ResourceMesh.cpp
```diff
@@ -2,11 +2,13 @@
 
 #include "Application.h"
 #include "CameraModule.h"
+#include "OpenGLModule.h"
 #include "ResourceMaterial.h"
 
 #include <Math/float2.h>
 #include <Math/float4x4.h>
 #include <SDL_assert.h>
+#include <chrono>
```
IDarkgenesis, line 11: Instead of using chrono here, shouldn't this be inside the Engine timer for the high resolution timer?

SirKaza: Right now, EngineTimer is using SDL_GetTicks() which is not as precise as chronos high_resolution clock so the elapsed time between before rendering mesh and after will be in floats equals to 0 seconds. Might be a good idea inside EngineTimer having what I am using here so we can reuse this code. 🤔

caspi99 (approved):
👍

PR #53, opened 2025-03-11
Fixed bug with wrong resource reference count

LukasGanter:
When a resource is created and returned its reference count was not incremented. That led to the issue that is described in the bug report


SirKaza (approved):
Nice 👍

PolCurto (approved):
Nice

PR #54, opened 2025-03-12
THO-216 library integration freetype

caspi99:
Adding library to the engine. If you open the engine, in the console you can see a font loading as a test of the library working


SirKaza (changes requested):
There are both in debug and release builds like 30 warnings like these:

"warning LNK4099: PDB '' was not found with 'freetype.lib(ftinit.obj)' or at ''; linking object as if no debug info"

SirKaza (changes requested):
Still warning links. There is warning 4099 and one 4098.

SirKaza (approved):
Nice change

PolCurto (approved):
Tested both Debug and Release, no warnings or anything. Seems good 👍

PR #55, opened 2025-03-12
Task/Skinning-THO-147

PolCurto:
To Test:
There is an already saved scene with loaded models. The zombunny and the player have skinning, you can move bones around and see what happens. Test meshes and models importing / loading, scene saving and loading with objects that have skinning, it's where I think is most probable that some bugs appear.

IMPORTANT: With skinning, now importing meshes collects two more vertex data: Joints and Weights. Delete all old meshes, as it is possible that some old meshes give problems or directly crash the engine (it shouldn't but you never know). From now on all meshes are imported taking that into account, so there should be no problems with future meshes, only old ones.

There is still the option to display the model skeleton (it is known there is always one line going to the scene origin when displaying that).

Also, to have models independent of the library, when importing a model it is saved in an assets folder called /Models with the .gltf name, and in the Library with the UID name. A meta file is created, and if you delete the file from library, when opening the engine the model in assets is directly copied to the library with the UID as a name, because they are the same file.


DHazardFenrir (approved):
The skinning works as expected, I could move the bones with the gizmo and the meshes deforms. I also could input manual values and deforms as well. The only thing is that the scene is quite dark, so I had difficulties to see the bone changes considering player and zombunny have very dark colors on the texture. 

Great job on the model reimporting when deleted, tested and worked like a charm.

LukasGanter (approved):

PR #59, opened 2025-03-14
Character motion component tho 198

MarcLopezS:


LukasGanter (changes requested):
The component works as expected, very good job!
Some small adjustments: 
- Missing overrides in CharacterControllerComponent (Save, Destructor, ...)
- GetTargetDirection() returns value instead of const ref
- Instead of saving the deltaTime in the sceneModule, please add the deltaTime variable to the base components render function and pass it that way to keep it consistent with the rest of the application
- When moving out of the viewport the movement stops working (No need to fix this as these GOs need to be removed from the octree anyway)
- Format the CharacterControllerComponent files according to our format rules

SobrassadaEngine/Scene/Components/ComponentUtils.h
```diff
@@ -20,16 +20,18 @@ enum ComponentType
...
-    LAST  = COMPONENT_DIRECTIONAL_LIGHT
+    LAST  = COMPONENT_CAMERA
 };
 
 static const std::unordered_map<std::string, ComponentType> standaloneComponents = {
     {"Mesh",              COMPONENT_MESH             },
     {"Point Light",       COMPONENT_POINT_LIGHT      },
     {"Spot Light",        COMPONENT_SPOT_LIGHT       },
     {"Directional Light", COMPONENT_DIRECTIONAL_LIGHT},
+    {"Character Controller", COMPONENT_CHARACTER_CONTROLLER}
```
DHazardFenrir, line 34: Add a coma after COMPONENT_CHARACTER_CONTROLLER} because if not it makes erros pop up

SobrassadaEngine/Modules/SceneModule.h
```diff
@@ -98,10 +98,14 @@ class SceneModule : public Module
     
     Octree* GetSceneOctree() const { return loadedScene != nullptr ? loadedScene->GetOctree() : nullptr; }
 
+    float GetCurrentDeltaTime() const { return currentDeltaTime; } 
+
+private:
     void SetMainCamera(CameraComponent* camera) { loadedScene->SetMainCamera(camera); }
```
DHazardFenrir, line 104: SetMainCamara Change it to public, since you are using it outside the class when you are calling GetSceneModule (one example of this is line 50 on Camera Component.cpp)

DHazardFenrir (changes requested):
Also fix merge errors with vcxproj and filters, but everything else is good! good work!

LukasGanter (approved):
Formatting is still missing, but still approved

DHazardFenrir (approved):
Everything good!

PR #60, opened 2025-03-14
THO-73 class setup for lerp

caspi99:
For testing (For example):
```
void TestInterpolation() {
    float a = 0.0f, b = 10.0f;
    float f1Lerp = Interpolation::Lerp(a, b, 0.5f);
    GLOG("Lerp float(0, 10, 0.5): %f", f1Lerp);

    float3 v1(0, 0, 0), v2(10, 10, 10);
    float3 f3Lerp = Interpolation::Lerp(v1, v2, 0.5f);
    GLOG("Lerp float3(0,0,0 -> 10,10,10, 0.5): (%f, %f, %f)", f3Lerp.x, f3Lerp.y, f3Lerp.z);

    Quat q1 = Quat::identity;
    Quat q2 = Quat::RotateAxisAngle(float3(0, 1, 0), 90.0f * (M_PI / 180.0f));
    Quat qLerp = Interpolation::Lerp(q1, q2, 0.5f);
    float3 euler = qLerp.ToEulerXYZ() * (180.0f / M_PI);
    GLOG("Slerp quaternion(Identity, 90Y, 0.5): (%f, %f, %f) degrees", euler.x, euler.y, euler.z);
}
```


PolCurto (approved):
Tested it both on Debug and Release and works fine. 
If you lerp a weird type like a string it gives a compilation error. I don't know if certain types can be forbidden with a templated function like this, but either way we will use it for numerical values so I think it's fine.

DHazardFenrir (changes requested):
Please solve filter erros.

DHazardFenrir (approved):
Good work! everything works as expected, conflicts with main resolved, and it was investigated if we could delimit the template with certain types but after test it goes crazy with mathgeolib types and the only solution would be denying 1 by 1 which can't be used.

PR #61, opened 2025-03-14
Task/Resource prefab THO-139

PolCurto:
HOW IT WORKS: To create a prefab, right click in a gameObject in the hierarchy and select "Create Prefab". In the name it will appear the prefab UID (ugly but useful for now). In the same menu used to load models, now you can load prefabs too. Loading a prefab will create a new gameObject with all the children and components saved. 
When right-clicking a prefab, instead of "Create Prefab" the button will say "Update prefab". When clicking this, the prefab file will be updated, and all prefab instances in the scene are also updated while staying in place. Instances of the prefab created in other scenes will also updated.

TO TEST: Basically create combinations of gameObjects and components, create prefabs from them, instantiate them, update them with other instances of the prefab in the scene and save/load scenes with the prefabs to see if they are properly loaded/updated and see if it crashes. I've tested many combinations without them crashing, but you never know which will be the one. There is already an scene with prefabs in it.

BUGS FIXED: There were 2 bugs in main now fixed here:
1. When loading a scene with 2 or more instances of the same light type, only one of them worked.
2. When deleting a gameObject with 3 or more children per level, some of them were not removed from the gameObjectsContainer. I don't know why exactly, but changing the method RemoveComponentHierarchy() from recursive to iterative has fixed it.

KNOWN ISSUES: 
1. As with the other resources, creating a prefab with the same name as another one replaces the last one in assets because their name is the same, so only one prefab with the same name can exist for now. 
2. For now prefabs are always created at the root level, so if a prefab is child of a gameObject, if it is updated it will move to the root level again. This should be changed, but I think better do it in the Prefab Editor task and merge this for now.

NOTE: Now Component has a pure virtual method Clone(), which has to be implemented with every Component type. Just a reminder for when new components are added.


LukasGanter (changes requested):
Basic functionality is there, looks amazing!
Some small issues: 
- AABBs for meshComponents are not calculated correctly (Meshes AABB is not used, instead only the default component AABB is transformed)
- In the example prefab, the zombunny does not render its material for me (In the example scene the materials are either messed up or are not rendering at all

MarcLopezS:
Basic funcionality works perfectly. Good Job! :)

I've found something that I don't know if it has been thought or will be implemented in the future. When I save a Scene with a  prefab being a child of another prefab, and then I load the saved scene, the prefabs are separated and are no longer nested. 

It also happens to me that the material is not rendering in zombunny in the example Scene.

MarcLopezS (changes requested):
Try to look how it can be saved the data of a prefab as a child of another prefab and, when load, show correctly this case.

MarcLopezS (changes requested):
I didn't saw that you metioned it in the comments. It's a similar case the one you contemplated and the one I found related when the prefab is updated (it will work the same when it is saved and load it). 

Check Lukas requested changes to proceed, but apart from that, for me it's good to go.

Let us know when it's updated for retesting again.

PolCurto:
I've fixed some things regarding the merge with skinning, and the AABB problem should be fixed. Regarding the materials bug, I'mm not sure if it is exclusive to this branch, for now it has not happened to me again. 

There is again an scene with prefabs and some prefabs already created, play around with them and see what happens.

As commented, the issue with prefabs parents will be fixed in the prefab editor task, as well as other details that may come up

MarcLopezS (approved):
I've tested again and now it renders correctly the materials when loading scene. I've tested diferent combinations on prefabs and seems good to me.

There is a problem with loading multiple times the scene sometimes, but is not related with this case, so you can ignore this for this task.

Everything seems to be working. Good job! :)

PR #63, opened 2025-03-15
Go parent be children THO-181

MarcLopezS:
This is the fix of the bug found when a GameObject parent can be child of it's children and ends in an infinite loop. This bug has been fixed. In order to test it, I tried to create GameObject parents and children and try to drag and drop a parent to their own children. Also has been tested that the functionality drag and drop hasn't been afected to the other cases of children and parents.


caspi99 (approved):
It solves the error 👍 🥒

IDarkgenesis (approved):
The loop is finished 🕳️

PR #65, opened 2025-03-16
Fixed Library-graph-vis-THO-195

DHazardFenrir:
This is to add the graphic node library for the state machines. Could also be used to make visual scripting. 
I used a different library than the one they showed us in class, but I talked with the teacher and I got it greelight for using it on your project. 

This also works on release.

How to test:
-Go to Window->Engine Editor Window->Nodes Engine Editor. This should open a grid. You can zoom in, zoom out, move nodes.
-Right click on the mouse button to add a node. Right now you can only sum two numbers in nodes. 

BEHAVIOURS TO BE AWARE OF. NOT BUGS:

-If you put two grids overlapping, the program will take this as a 'shared' grid since they are on the same screen coordinates. So if you add a node here, and then move it to the other grid, it will move. If you add a node and its on the middle of the edge between two grids, the node will duplicate if you move one grid. 

-If you put to grids on the same anchor point (in the middle anchor point) in a way that is two tabs next to each other in the same space, it will take it as a shared space as well, meaning that the nodes you put on one side will clone on the other side, and if you connect one node on one side it will show in the other as well. If you separate the grids, this won't be any longer and will be taken as a separate space.  Here are some gifs that show it. 
OtherCases
Case1

SIDENOTE: I needed to create another new branch from main and delete the other to fix everything that was broken and prevented the build. Right now this is up to date to main, the filters are fixed, and you can run it in debug and release. The behavior is the same as the other one.


PolCurto (changes requested):
It works fine in both debug and release. Just include the license in the proper folder and the about window and is good to go

DHazardFenrir:
license and about updated

PolCurto (approved):
Nice 👍

SobrassadaEngine/UI/EngineEditors/Nodes/CustomNode.h
```diff
@@ -0,0 +1,38 @@
...
+
+    void draw() override
+    {
+      
+        ImGui::SetNextItemWidth(100.f);
+        ImGui::InputInt("##ValB", &m_valB);
+
+         
+         ImGui::Text("Resultado: %d",getInVal<int>("IN_VAL") + m_valB);
+
```
caspi99, line 24: Why a lot of blank spaces here

SobrassadaEngine/UI/EngineEditors/Editor/NodeEditor.cpp
```diff
@@ -0,0 +1,63 @@
...
+                            myGrid->addLink(newLink);
+                        }
+                    }
+                }
+            }
+        }
+    );
+    ImVec2 currentSize = ImGui::GetWindowSize();
+    if (currentSize.x < 150 || currentSize.y < 150)
+    {
```
caspi99, line 57: Here you need to delete the brackets

caspi99 (changes requested):
Just change the format and its good to merge

DHazardFenrir:
Formatting updated I will solve the conflicts with main.

DHazardFenrir:
Merged with main

caspi99:
Now its broken 🙃 
image

DHazardFenrir:
I fixed some duplicate errors (that was broke it) run on release

caspi99 (approved):
:+1: ♿

PR #66, opened 2025-03-16
THO-221 Dynamic data structure (FPS Patch)

IDarkgenesis:
Added dynamic data structure (right now a quadtree).

Trees are now being updated as least as possible inside the postupdate function in scene module so no more constant updtades while dragging ImGuizmo.

Changed Raycastcontroller to work with multiple data structures for raycasting.

To test: load the xd.scene or create a new one and add at least 5 elements and try switching between dynamic and static to check if trees update properly, same with object picking with game object with meshes.

Known issue: Closing scene with some boxes or tree rendering active it creashes, this is being fixed on: THO-219


SirKaza (approved):
Good changes! Works as expected, should have a task to propagate the dynamic to game object children's in a near future.

caspi99 (approved):
Good job! Works as expected 👑

PR #69, opened 2025-03-16
Bug/some bugs fixed THO-219

SirKaza:
I fixed some bugs like orbiting and disabling gizmo or not tabing when zooming the scene.
I tried to fix the stop play mode bugs, but I can't figure out what is happening for this vertical slice.

How to test: try orbiting and u can't go to infinite values while moving the gizmo, or try zooming with Left Alt and right click and the window will not be tabbed.


caspi99 (approved):
It seems all good. Good job!

LukasGanter (approved):
Great work

PR #71, opened 2025-03-18
Fixed memory leaks in EngineConfig and ComponentUtils - THO-224

LukasGanter:
Fixed two memory leaks. 
- EngineConfig did not get deleted
- static list in ComponentUtils showed as memory leak, according to Marc its an issue in the vs debugger. Still removed it though


SirKaza (approved):
Nice changes. Tested and works properly.

PR #72, opened 2025-03-18
Bug/problem with textures THO-230

caspi99:
image

Testing: Don't use an intel GPU, they're stupid as fuck


caspi99:
I'm not 100% sure, but If someone wants to know, I think integrated GPU's does not support bindless textures because integrated GPUs does not have an integrated ram

SirKaza (approved):
Good changes! Tested and works properly. I had to delete library files to render things as expected, I don't know why.

LukasGanter (approved):

PR #73, opened 2025-03-19
State machine resource - THO-196

PabioGF:
I'm at work so I can't really test it deeply but it seems to work fine. Have in mind that this was done in a rush for the VS1 so it's not the final version, there will be changes in the future when we implement the editor, I'm sure. 

Please test that there's not errors that i could have missed before merging to main, Thanks!


SobrassadaEngine/FileSystem/StateMachineManager.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include "FileSystem.h"
+#include "ResourceManagement/Resources/ResourceStateMachine.h"
```
SirKaza, line 4: forward declaration

SobrassadaEngine/ResourceManagement/Resources/ResourceStateMachine.h
```diff
@@ -0,0 +1,82 @@
...
+    const Clip* GetClip(const std::string& name) const;
+    const State* GetState(const std::string& name) const;
+    const Transition* GetTransition(const std::string& fromState, const std::string& toState) const;
+
+    bool ClipExists(const std::string& clipName) const; 
+
+    void SetUID(UID uid) { stateMachineUID = uid; };
+
+  private:
+    UID stateMachineUID;
```
SirKaza, line 80: Resource base class already has an UID, so not needed here

SirKaza (approved):
Good work with the Hash String.
Try to take a look to formatting and coding style in the includes (order).
Not sure as we talked if state machines needs to be saved in binary or in rapidjson.

LukasGanter (approved):

PR #74, opened 2025-03-19
Texture editor THO-153

jmartiolivart:
Before creating the pull request, here’s a summary:
- Implemented the Texture Library Editor to list and display loaded textures.
- Added columns for name, width, height, and memory size in KB.
- Fixed issues with texture preview display, ensuring textures render correctly when selected.
- Addressed some errors related to ResourceTexture and ImTextureID compatibility.

The task is not yet fully finished, but I will create the PR now to avoid merge conflicts. I have tested the changes, and I don’t believe they break anything in the main branch. However, let me know if you notice any unexpected issues. I will continue working on a new branch after the merge.

### DIFF
--- a/SobrassadaEngine/UI/EngineEditors/Editor/TextureLibraryEditor.cpp
+++ b/SobrassadaEngine/UI/EngineEditors/Editor/TextureLibraryEditor.cpp
@@ -1,0 +1,70 @@
+#include "TextureLibraryEditor.h"
+#include "Application.h"
+#include "LibraryModule.h"
+#include "ResourceManagement/Resources/ResourceTexture.h"
+#include "imgui.h"
+#include <TextureImporter.h>
+
+TextureLibraryEditor::TextureLibraryEditor(const std::string& editorName, UID uid) : EngineEditorBase(editorName, uid)
+{
+}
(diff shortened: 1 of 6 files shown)


SobrassadaEngine/UI/EngineEditors/Editor/TextureLibraryEditor.cpp
```diff
@@ -0,0 +1,70 @@
...
+    ImGui::NextColumn();
+    ImGui::Text("Height");
+    ImGui::NextColumn();
+    ImGui::Text("Memory Size (KB)");
+    ImGui::NextColumn();
+    ImGui::Separator();
+
+    for (const auto& [name, textureUID] : textureMap)
+    {
+        ResourceTexture* texture = TextureImporter::LoadTexture(textureUID);
```
SirKaza, line 49: Here should be calling Resource Module Request Resource method so if resource is loaded it is just returning the resource. I'm not sure if here it should be loading all the textures that are metadata into memory.

SirKaza (approved):
When I try to open the Texture Library window from the menu bar, it crashes the engine.

PR #79, opened 2025-03-26
Update keys mov character controller tho 324

MarcLopezS:
Added movement in X-axis as a request from the art team. Now keys "A" and "D" makes horizontal movements.

To mahe coherence with rotation, now rotation keys are binded with "Q" (rotate left) and "E" (rotate right).

In order to test this, create a gameObject with a mesh component as a child and the parent should have the character Controller component. Once in play mode, play with WASDQE keys.

NOTE: Has been found a bug when play button is pressed and afterwards  stop button is pressed more than once, it unables to execute the engine. To solve this, you are forced to delete the save scene that was using. It is unknown if this bug has been solved in VS2 branch.


IDarkgenesis (approved):
Thanks for being this fast, good job

PR #82, opened 2025-04-01
Scripting setup THO-312 

caspi99:
For testing:
You can create an script component and add the rotation script and a mesh. You can change the rotation velocity as I do in the video and compile again without closing the engine (You can close it, but I implemented hot reloading 🔥🔥🔥) . You can save and load the scene and click the play button to see the script doing things.

https://github.com/user-attachments/assets/01c68113-39d6-421c-9999-5bf85802a58c


SobrassadaEngine/Scene/Components/ScriptComponent.cpp
```diff
@@ -0,0 +1,108 @@
...
+                    if (scriptInstance != nullptr) DeleteScript();
+                    CreateScript(scriptType);
+                }
+            }
+            ImGui::EndPopup();
+        }
+    }
+}
+
+void ScriptComponent::CreateScript(std::string scriptType)
```
IDarkgenesis, line 97: String value passed by copy instead of reference

SobrassadaEngine/Scene/Components/ScriptComponent.h
```diff
@@ -0,0 +1,31 @@
...
+    void Clone(const Component* other) override;
+
+    void Update(float deltaTime) override;
+    void Render(float deltaTime) override;
+    void RenderEditorInspector() override;
+
+    void CreateScript(std::string scriptType);
+    void DeleteScript();
+
+    std::string GetScriptName() const { return scriptName; }
```
IDarkgenesis, line 24: String name passed by copy instead of reference

IDarkgenesis (changes requested):
Good job ! Please have a look at the comments

IDarkgenesis (approved):
👎

PR #83, opened 2025-04-03
Default_material_mesh_THO-244

MarcLopezS:
Added funcionality to have each metamesh it's default material ready in it's data. Also, when a mesh is selected, automatically assigns the default material to the mesh, updating the field "Select Material".

In order to test it:
- While engine open, drag and drop a model gltf in the engine.
- Create a GameObject and add a Mesh component.
- Select the imported mesh.

When doing this steps, you should see that the material is autoassigned at first stance. 
Test various models to check if there is any problem with some specific ones.

NOTE: the default models (cone, cube, sphere,...) are being modified with the default material. For the basic meshes has been done purposely like that way to make coherence. To display the pink texture to notice the user that the material is not found when selecting the mesh, will be implemented in the task THO-243, with the implementation of the default texture.


LukasGanter (dismissed):
When loading a mesh in a scene, the default material will always be loaded before a check is done if this mesh has already a custom material assigned. So if a mesh and material is assigned to a mesh component, the default material will be loaded within the AddMesh function and deleted immediatelly after in the AddMaterial function. Maybe add a check inside the MeshComponent whether it has a material assigned, and if so don´t load the default.

Also, at least for me, when starting the release build of the engine outside the IDE no meshes are rendered at all. Not sure if this happens only on this branch.

LukasGanter (commented):
Regarding the meshes which are not visible, this doesn´t work on VS2 either.

MarcLopezS:
Added feature that @LukasGanter asked when selecting a mesh, if no material was selected, it will use the predefined one in the metaMesh. Otherwise will remain using the current one.

SobrassadaEngine/FileSystem/Importers/SceneImporter.cpp
```diff
@@ -66,7 +64,10 @@ namespace SceneImporter
                 {
                     matUID = matIndices[matIndex];
                 }
-
+                
+                UID meshUID      = MeshImporter::ImportMesh(model, srcMesh, primitive, name, filePath, targetFilePath, INVALID_UID, matUID);
```
PolCurto, line 68: can be const

SobrassadaEngine/Scene/Components/Standalone/MeshComponent.cpp
```diff
@@ -186,6 +186,13 @@ void MeshComponent::AddMesh(UID resource, bool updateParent)
         App->GetResourcesModule()->ReleaseResource(currentMesh);
         currentMeshName    = newMesh->GetName();
         currentMesh        = newMesh;
+        
+        if (currentMaterial == nullptr)
+        {
+            UID defaultMat = newMesh->GetDefaultMaterialUID();
```
PolCurto, line 192: also can be const

PolCurto (approved):
Good job, works as expected. When the material is already loaded, changing the mesh for another one doesn't update the material, but reading the comments it seems to be intended. Just check the two missing const

caspi99 (approved):
Approved! Kurwa accepted! 🏉

PR #86, opened 2025-04-04
bug/gizmo window tab THO-319

SirKaza:
Things done:
· Small fix on models.
· Deleted programs uint from resource module.
· Fixed bug that gizmo was being displayed even if scene window was tabbed.

How to test:
Try tabbing windows in scene window and gizmo should not be displayed.


PolCurto (approved):
Nice

caspi99 (approved):
Funca 🔥

PR #87, opened 2025-04-04
Default texture material & Change texture in material ( THO-243 and THO-242)

MarcLopezS:
Tasks THO-242 (Andreu Castaño) and THO-243 (Marc López) have been implemented and merged together. 

The implementation made is that the default texture UID is now registered in the MetaMaterials and saved in the Resource Material. Now. whenever a material does not found a texture, the FALLBACK TEXTURE will be added by default, letting the user know that there is a problem in the texture with that material.

It is difficult to test this, because it should be required a material that has broken textures or a material that does not have a texture itself. Will be needed a model that contains a material with those characteristics.

For the other task, have been implemented the selectables of the Diffuse Texture and the Metallic Texture. Andreu and I both tested and worked fine for us. Also, if a model does not have a Metallic texture, the UI button adapts its position to not confuse the user.

To Test:
- Check that all the buttons work propertly
- Try to select the different textures for the same material and check that renders the texture propertly.

Also, as I mentioned before, if we have a model that has a material that does not have a diffuse texture, should appear automatically the material with the FALLBACK TEXTURE. It can't be tested until we have a model with those characteristics. (already asked the artist to have a model like that).


LukasGanter (changes requested):
These tasks should update the meta data files for all the engine meshes as well. 
That would also make it easier to test everything. 
Also, please make sure that all files generated by earlier engine version work with the new system as well. For now, when i.e. a material meta file is missing, the engine doesn´t render the default texture. The mesh just disappears.

LukasGanter (approved):
Everything works well now, good job.

SobrassadaEngine/ResourceManagement/Resources/ResourceMaterial.cpp
```diff
@@ -71,7 +149,61 @@ void ResourceMaterial::OnEditorUpdate()
...
+{
+    if (newTexture == INVALID_UID) return NULL;
+
+    ResourceTexture* texture = TextureImporter::LoadTexture(newTexture);
+    if (texture != nullptr)
+    {
+        glMakeTextureHandleNonResidentARB(textureGPU);
+        glDeleteTextures(1, &textureToChange.textureID);
+
+        UID handle = glGetTextureHandleARB(texture->GetTextureID());
```
PolCurto, line 191: can be const

PolCurto (changes requested):
When clicking the "Select material" button, the pink material is immediatly loaded, and then selecting the default material also displays the pink texture.

SobrassadaEngine/Scene/Components/Standalone/MeshComponent.cpp
```diff
@@ -139,14 +139,17 @@ void MeshComponent::RenderEditorInspector()
...
         }
 
-        if (ImGui::IsPopupOpen(CONSTANT_TEXTURE_SELECT_DIALOG_ID))
+        if (ImGui::IsPopupOpen(CONSTANT_MATERIAL_SELECT_DIALOG_ID))
         {
-            AddMaterial(App->GetEditorUIModule()->RenderResourceSelectDialog<UID>(
-                CONSTANT_TEXTURE_SELECT_DIALOG_ID, App->GetLibraryModule()->GetMaterialMap(), INVALID_UID
-            ));
+
+            UID chosenMatUID = App->GetEditorUIModule()->RenderResourceSelectDialog<UID>(
```
PolCurto, line 148: one last const 🍶

PolCurto (approved):
Works properly now, good job!

PR #88, opened 2025-04-04
Task/resource font THO-325

PolCurto:
Added resource font. Fonts can be imported as any other file, drag & drop or import window. Fonts are .ttf (you can download any font from internet with TrueType format). When imported, you can assign the font to a label component and the text will use it.

TO TEST:  Import different .ttf files and assign fonts to labels, try saving and loading them and see if something breaks.


SobrassadaEngine/FileSystem/Importers/FontImporter.cpp
```diff
@@ -0,0 +1,56 @@
+#include "FontImporter.h"
+
+#include "Application.h"
+#include "FileSystem.h"
+#include "LibraryModule.h"
+#include "MetaFont.h"
+#include "ResourceManagement/Resources/ResourceFont.h"
```
srLlobet, line 7: Should be "ResourceFont.h"

srLlobet (changes requested):

SobrassadaEngine/FileSystem/Importers/FontImporter.cpp
srLlobet: I tried to load a weird font with pictures of plants and it worked! Good job!

srLlobet (approved):

SobrassadaEngine/FileSystem/Importers/FontImporter.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include "MetaFont.h"
+#include "ResourceFont.h"
+
+namespace FontImporter
+{
+    UID ImportFont(const char* filePath, const std::string& targetFilePath)
+    {
+        // Copy font to assets folder
+        const std::string relativePath = ASSETS_PATH + FileSystem::GetFileNameWithExtension(filePath);
+        std::string copyPath           = targetFilePath + relativePath;
```
caspi99, line 15: const

caspi99 (approved):
Good job! Just make this small change and its good to go 🎲

PR #89, opened 2025-04-05
Task/game UI anchor&pivots THO-317

PolCurto:
DESCRIPTION: Added anchors mimicking unity functionalities. Modifying the transform2D position updates the global transform and vice versa, taking into account pivots and anchors, this way the UI can be moved around uisng the gizmos which makes it more comfortable. 
You can detach the anchors (max and min X, max and min Y). When an anchor axis is detached, the transform2D inspector changes and instead of the position and the size, for that axis you can modify the margins, which are the distance to the detached anchors.
If the anchors are all in the same place, in the inspector you can modify the positions and widht/height, which will take into account the distance to the anchor and will move along if the parent is modified.

TO TEST: Create Labels inside a Canvas and move the anchors, create children and modify them as well, move the pivots, etc. Basically move things around and test the anchors with children inside. I'm pretty sure it will eventually break.

Labels can be created outside a Canvas, but are not rendered and Transform2D can't be used.


LukasGanter (changes requested):
System works, looks great!
Issues i found so far:
- Trying to delete the transform2D component crashes the engine
- When moving the pivot to (1, 1) the label still rotates around (0, 1)

PolCurto:
Both problems solved. I noticed that if you rotate a label, its children will do weird things with deatached anchors, but I don't think that's worth the effort to solve

IDarkgenesis (changes requested):

PolCurto:
Fixed the warnings. The anchors of a widget get updated when the parent is resized. For example, if the anchors are in the two horizontal extremes and the parent width is increased, the anchors move along them. Regarding the last image, the function can't be const because it's not a class, it's a namespace.

https://github.com/user-attachments/assets/ea6015a3-9e16-431d-8942-3cf49901bbaa

IDarkgenesis:
> Fixed the warnings. The anchors of a widget get updated when the parent is resized. For example, if the anchors are in the two horizontal extremes and the parent width is increased, the anchors move along them. Regarding the last image, the function can't be const because it's not a class, it's a namespace.
> 
>  Grabacion.de.pantalla.2025-04-10.162044.mp4

Sorry I'm stupid, I meant static for the function

PolCurto:
Added static

IDarkgenesis (approved):
Good job!

🌵

LukasGanter (approved):
All issues adressed, great work!

PR #93, opened 2025-04-08
Version 0.3.1

IDarkgenesis:
Fixes a linking error when compiling the game and fixes an engine freeze when moving collider components.


SirKaza (approved):
Tested in all the configurations and works. There are warning in debug and game debug about pdb of physics libraries.

PR #97, opened 2025-04-09
THO-334 Multiple Selection

IDarkgenesis:
Adds the multi selection feature.

To test:
Add different game objects / models to the scene and by holding Left Shift and clicking objects AABB should render arround the selected ones and by moving the gizmo all objects should move together.

I know that those objects dissapear temporarily from the tree node view (code limitations).

Clicking outside the scene window or just left clicking should restore the tree view and all multiselected objects should stop rendering the AABB.

multiSelect

### DIFF
--- a/SobrassadaEngine/Scene/Scene.cpp
+++ b/SobrassadaEngine/Scene/Scene.cpp
@@ -654,7 +654,7 @@ void Scene::AddGameObjectToSelection(UID gameObject, UID gameObjectParent)
         GameObject* selectedGameObject       = GetGameObjectByUID(gameObject);
         GameObject* selectedGameObjectParent = GetGameObjectByUID(gameObjectParent);
 
-        selectedGameObjectParent->RemoveGameObject(gameObject);
+        //selectedGameObjectParent->RemoveGameObject(gameObject);
 
         multiSelectParent->AddGameObject(gameObject);
 
--- a/SobrassadaEngine/Scene/Scene.cpp
+++ b/SobrassadaEngine/Scene/Scene.cpp
@@ -303,7 +303,7 @@ update_status Scene::Render(float deltaTime)
     {
         GameObject* gameObject = GetGameObjectByUID(gameObjectIterator.first);
 
-        AABB aabb              = gameObject->GetHierarchyAABB();
+        const AABB aabb              = gameObject->GetHierarchyAABB();
         
         for (int i = 0; i < 12; ++i)
             debugDraw->DrawLineSegment(aabb.Edge(i), float3(1.f, 1.0f, 0.5f));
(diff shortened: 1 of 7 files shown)


SobrassadaEngine/Scene/Scene.cpp
```diff
@@ -644,6 +718,17 @@ const std::vector<Component*> Scene::GetAllComponents() const
...
 }
 
+UID Scene::GetMultiselectUID() const
+{
+    return multiSelectParent->GetUID();
+}
+
+void Scene::SetMultiselectPosition(const float3& newPosition)
+{
+    float4x4 localMat = float4x4::FromTRS(newPosition, float4x4::identity, float3::one);
```
PolCurto, line 728: can be const

SobrassadaEngine/Scene/Scene.cpp
```diff
@@ -293,6 +297,18 @@ update_status Scene::Render(float deltaTime) const
         }
     }
 
+    DebugDrawModule* debugDraw = App->GetDebugDrawModule();
+
+    for (auto& gameObjectIterator : selectedGameObjects)
+    {
+        GameObject* gameObject = GetGameObjectByUID(gameObjectIterator.first);
+
+        AABB aabb              = gameObject->GetHierarchyAABB();
```
PolCurto, line 306: can be const

PolCurto (commented):
Tested it and works fine with root gameObjects. However you can't add a child mesh to the multiselection because the parent is always selected, and the artists will probably complain about that 💀
 
I don't know if it's fine like this for now?

IDarkgenesis:
> Tested it and works fine with root gameObjects. However you can't add a child mesh to the multiselection because the parent is always selected, and the artists will probably complain about that 💀
> 
> I don't know if it's fine like this for now?

Multiselection always picks the top most parent because of models, if you want to stop beforehand there is the option to mark a gameObject as Top parent, and the picking will stop and select that one.

PolCurto (approved):
Okay, just checked it works fine. Good job!

LukasGanter (changes requested):
Works well in the scene, great job
- the multi selection should also be visible inside the game objects hierarchy, also it should be possible to select multiple game objects via the hierarchy

IDarkgenesis:
> Works well in the scene, great job
> 
> * the multi selection should also be visible inside the game objects hierarchy, also it should be possible to select multiple game objects via the hierarchy

I have made it so game object do not dissapear from the hierarchy but in order to mantain the selected game objects tree nodes selected it would require to hold an array or vector with as many positions as nodes in the scene to hold if they are selected, I think is too much to handle that, eacg node requires the flag to know if its selected.

If we allow to select from the hierarchy it introduces to many conditions to handle, relese when going in play mode, what to do when deleting items when multiselected, ... too much of a time investment for a code that wasn't created for that feature.

LukasGanter (approved):
Approved after testing with artists

PR #98, opened 2025-04-10
bug/ModelsFixed

SirKaza:
Things done:
· Models import is iterative.
· Changed how import works.
· Reading joints type, which could be different to indices.

How to test:
Import different GLTFs and test if results are expected.


IDarkgenesis (changes requested):
Reimporting the Main character and Zoombuny in release and debug mode freezes the engine when loading the model

IDarkgenesis (approved):
Amazing job fixing the model loading thanks!

🪨

PR #99, opened 2025-04-11
task/Moved Optick from Release to Profile config

SirKaza:
If someone downloads releases builds, they don't need to have Visual Studio installed to run .exe
Optick will be available in Profile and GameDebug configs now.

How to test:
Run Profile config and check that Optick works in this config.
Run Release and Optick shouldn't work.


LukasGanter (approved):
Build works correctly on a pc without an installed version of visual studio. 
Great work fixing it so fast!

srLlobet (approved):
Looks good! Optick is properly disabled in release and enabled in profile. Good job!

PR #100, opened 2025-04-11
Fix vcxproj

srLlobet:
Also noticed recast was included twice in debug.

All branches should work now


PolCurto (approved):
Works fine now 🦺

SirKaza (approved):
:shipit:

PR #101, opened 2025-04-11
Bug/Prefab crash on load THO-366 & NavMesh crash empty scene THO-369

PolCurto:
The following bugs have fixed:
- Loading a scene with prefabs that contain meshes crashes the engine
- Loading a prefab does not apply the skinning
- Overriding a prefab makes them appear in the origin

Check if that's true, maybe better with a prefab that contains some artists models

I have also added to this PR the crash when creating a NavMesh in an empty scene or when a meshComponent had no mesh attached, to not create another PR for 2 lines of code. Test that aswell


LukasGanter (approved):
Approved for both tasks, good job!

SirKaza (changes requested):
I tried to test with main char and zoombunny model prefabs and when loading the scene or loading a prefab it crashes in AnimationComponent Clone() in this line:
AddAnimation(otherAnimation->currentAnimResource->GetUID());

PolCurto:
Should be fixed now

srLlobet (changes requested):
Loading a prefab with animations doesn't load the animations.

Moreover, since the list of prefabs to load is only filled with non-repeatable names, if you make a zombunny prefab, and then make a cube and name it zombunny, it will overwrite the zombunny on the list. A suggestion would be to add some sort of counter to the name? This is important because if you make a bunch of unnamed gameobject prefabs, only the last one will be saved.

Apart from this, great job! A lot of the bugs are fixed!

PolCurto:
Added a "(1), (2), ..., (n)" Suffix to prefabs with repeated names.

Regarding the animations problem, I was aware of that, but it's not part of this task

SobrassadaEngine/FileSystem/PrefabManager.cpp
```diff
@@ -32,7 +32,25 @@ namespace PrefabManager
             override ? gameObject->GetPrefabUID() : App->GetLibraryModule()->AssignFiletypeUID(uid, FileType::Prefab);
         savePath = App->GetProjectModule()->GetLoadedProjectPath() + PREFABS_LIB_PATH + std::to_string(finalPrefabUID) +
                    PREFAB_EXTENSION;
-        const std::string& name = gameObject->GetName();
+        std::string originalName = gameObject->GetName();
```
SirKaza, line 35: const &

SobrassadaEngine/FileSystem/PrefabManager.cpp
```diff
@@ -32,7 +32,25 @@ namespace PrefabManager
             override ? gameObject->GetPrefabUID() : App->GetLibraryModule()->AssignFiletypeUID(uid, FileType::Prefab);
         savePath = App->GetProjectModule()->GetLoadedProjectPath() + PREFABS_LIB_PATH + std::to_string(finalPrefabUID) +
```
SirKaza, line 33: Why define two times savePath?

SobrassadaEngine/Modules/ResourcesModule.cpp
```diff
@@ -117,7 +117,9 @@ void ResourcesModule::CreateNavMesh()
                 if (meshComponent)
                 {
                     const ResourceMesh* resourceMesh = meshComponent->GetResourceMesh();
-                    AABB aabb                        = gameObject->GetGlobalAABB();
+                    if (resourceMesh == nullptr) continue; // If a meshComponent has no mesh attached, ignore it
+
+                    const AABB aabb                        = gameObject->GetGlobalAABB();
```
SirKaza, line 122: & missing

SirKaza (changes requested):
Is crashing in when trying to load a scene with prefabs in line
Scene::RemoveGameObjectHierarchy() line 607
this->multiSelectParent was nullptr.

PolCurto:
Should be fixed

SobrassadaEngine/FileSystem/PrefabManager.cpp
```diff
@@ -26,17 +26,17 @@ namespace PrefabManager
...
         // Scene values
-        UID uid              = GenerateUID();
-        std::string savePath = PREFABS_LIB_PATH + std::string("Prefab") + PREFAB_EXTENSION;
+        UID uid = GenerateUID();
         UID finalPrefabUID =
             override ? gameObject->GetPrefabUID() : App->GetLibraryModule()->AssignFiletypeUID(uid, FileType::Prefab);
-        savePath = App->GetProjectModule()->GetLoadedProjectPath() + PREFABS_LIB_PATH + std::to_string(finalPrefabUID) +
-                   PREFAB_EXTENSION;
-        std::string originalName = gameObject->GetName();
+        const std::string& savePath = App->GetProjectModule()->GetLoadedProjectPath() + PREFABS_LIB_PATH +
```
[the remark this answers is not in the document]
PolCurto: oops, I'll fix that now and merge it

PR #103, opened 2025-04-12
Task/Game UI image THO-316

PolCurto:
Add the Image Component.

TO TEST: Similar to the label, the image must be child of a canvas to be rendered. You can change the source image and the color of it. By default the defaultTexture is applied. Try changing it, saving and loading and see if it works fine.

Also other UI-related bugs have been fixed in this PR (such as labels being drawn 3 times per frame or crashing when changing the hierarchy order of gameObjects with Transform2D).


LukasGanter (changes requested):
Under certain conditions the image is not rendered.
Steps to reproduce: 
- Add an image 
- Play and stop (twice)
- Image is not rendered anymore

SobrassadaEngine_XdEuUF5X64

PolCurto:
Should be fixed now

jmartiolivart (approved):
Tried everything from the PR description and also followed the steps mentioned about the image not rendering after playing/stopping twice works fine now. 
The image keeps rendering, texture and color changes apply correctly, and saving/loading keeps everything as expected. Didn’t run into any issues, good job 👍

LukasGanter (approved):
Bug fixed, good job

PR #104, opened 2025-04-12
Player movement update THO-212

MarcLopezS:
Now the player is able to walk though the navmesh, similar that the AI Agent component does. It clamps the position outside the borders of the navMesh in order to not fall and also in holes that could possibly have the navMesh inside.

To make things easier, a manual gravity has been applied in characterControllerComponent to control the height between the player and the navMesh. If we need to check this let me know and we can reformulate that part, but at the moment it works perfectly fine.

To test this:
- Create a Scene with some gameObjects that work as a floor and some obstacles
- Create some objects that can work as a height to be able to climb them. 
- Once the scene is created, create the navMesh and adjust the parameters according to the test you want to try.
- After NavMesh is created, place a gameObject that will act as the player with a mesh and a playerController component.
- You can place the player a little bit higher just in case it does not overlap the player. (the gravity will work and place in the correct spot in the navMesh)
- play with it with WASD keys.

If you find any error, please let me know to solve it as soon as possible.

Bug known that has been found: if you try to to on playMode with a gameObject that has a characterControllerComponent and the scene does not have a NavMesh, the engine crashes. This is known and I'm trying to sort it out to avoid crashing if a navMesh is not there.

PR #105, opened 2025-04-12
Player movement controller update 2 THO-212

MarcLopezS:
Now the player is able to walk though the navmesh, similar that the AI Agent component does. It clamps the position outside the borders of the navMesh in order to not fall and also in holes that could possibly have the navMesh inside.

To make things easier, a manual gravity has been applied in characterControllerComponent to control the height between the player and the navMesh. If we need to check this let me know and we can reformulate that part, but at the moment it works perfectly fine.

To test this:

Create a Scene with some gameObjects that work as a floor and some obstacles
Create some objects that can work as a height to be able to climb them.
Once the scene is created, create the navMesh and adjust the parameters according to the test you want to try.
After NavMesh is created, place a gameObject that will act as the player with a mesh and a playerController component.
You can place the player a little bit higher just in case it does not overlap the player. (the gravity will work and place in the correct spot in the navMesh)
play with it with WASD keys.
If you find any error, please let me know to solve it as soon as possible.

Bug known that has been found: if you try to to on playMode with a gameObject that has a characterControllerComponent and the scene does not have a NavMesh, the engine crashes. This is known and I'm trying to sort it out to avoid crashing if a navMesh is not there.

(Original pull request closed due to conflicts with merge. Original pull request #104 )

### DIFF
--- a/SobrassadaEngine/Scene/Components/Standalone/CharacterControllerComponent.cpp
+++ b/SobrassadaEngine/Scene/Components/Standalone/CharacterControllerComponent.cpp
@@ -5,6 +5,9 @@
 #include "GameObject.h"
 #include "InputModule.h"
 #include "SceneModule.h"
+#include "DetourNavMeshQuery.h"
+#include "ResourcesModule.h"
+#include "ResourceNavMesh.h"
 
 #include "Math/float3.h"
 #include "Math/float4x4.h"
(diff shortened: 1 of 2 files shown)


LukasGanter (approved):
Crash is fixed, everything works. Great job!

srLlobet (approved):
Looks very good and works! Good job!

PR #106, opened 2025-04-13
Animation Clips-THO-376

PabioGF:
Now the diferent clips in an animation are saved with a diferent name and sowed on the inspector of the component. 

You can select the diferent animations and play the one you selected.

I attach here the zombunny with 2 animations, that is the same animation copied 2 times. For seeing visually the diference between the original and the copy you will see that Take 002 has bones erased, so in this animation you will see for example that doesn't have legs. You can check also on the inspector the information of each one.
This bug is important for finishing the state machine task so please test it as soon as you can.

image

[ZomBunnySpecular (2) (1).zip](https://github.com/user-attachments/files/19726612/ZomBunnySpecular.2.1.zip)

### DIFF
--- a/SobrassadaEngine/Scene/Components/Standalone/AnimationComponent.cpp
+++ b/SobrassadaEngine/Scene/Components/Standalone/AnimationComponent.cpp
@@ -11,8 +11,8 @@
 #include "ResourcesModule.h"
 #include "SceneModule.h"
 
-#include "imgui.h"
 #include "Math/Quat.h"
+#include "imgui.h"
 
 AnimationComponent::AnimationComponent(const UID uid, GameObject* parent)
     : Component(uid, parent, "Animation", COMPONENT_ANIMATION)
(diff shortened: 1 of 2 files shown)


SobrassadaEngine/Scene/Components/Standalone/AnimationComponent.cpp
```diff
@@ -78,14 +78,20 @@ void AnimationComponent::OnResume()
...
 {
-
+    std::string originAnimation = "";
     if (resource != 0)
     {
         currentAnimResource = dynamic_cast<ResourceAnimation*>(App->GetResourcesModule()->RequestResource(resource));
-
+        const std::string animationName = App->GetLibraryModule()->GetResourceName(resource);
+        
+        size_t underscorePos            = animationName.find('_');
```
caspi99, line 87: const

SobrassadaEngine/Scene/Components/Standalone/AnimationComponent.cpp
```diff
@@ -78,14 +78,20 @@ void AnimationComponent::OnResume()
...
+        if (underscorePos != std::string::npos)
```
caspi99, line 88: Can be 1 line if

SobrassadaEngine/Scene/Components/Standalone/AnimationComponent.cpp
```diff
@@ -247,8 +253,36 @@ void AnimationComponent::OnInspector()
...
         ImGui::Text("Available Animations:");
+        const std::unordered_map<std::string, UID>& animationMap = App->GetLibraryModule()->GetAnimMap();
+
+        for (const auto& pair : animationMap)
+        {
+            const std::string& animationName = pair.first;
+
+            if (animationName.rfind(originAnimation, 0) == 0)
+            {
+                bool isSelected = (selectedZombunnyAnim == animationName);
```
caspi99, line 267: const

caspi99 (approved):
Solved bug. Good job! Just make the little changes and its good to go!

DHazardFenrir (approved):
Good job! tested and everything works!

PR #109, opened 2025-04-13
Camera component preview THO-235 and Materials without name THO-372

caspi99:
To test:
Add a camera and see the preview
There's a bug where we need to move the camera to render the camera but it works anyway (only in the Scene, in the play mode it works).
This bug exists in VS2 branch
Also solved the bug when a material does not have a name. You can check this gltf has no name in the material, import it and if you try to change the material it wont crash the engine. (You can try to import this gltf in other branch to see that if you try to change the material it crash the engine, but remember to reimport)


IDarkgenesis (approved):
Camera preview and fix to material not having names epic!

🎥

PR #110, opened 2025-04-14
Task/Game UI button THO-360

PolCurto:
Created the UI Button Component. When adding a button, it will automatically create an image component if there is none, and it will modify its color when hovered. You can change the colors and It can also be non-interactable, all this parameters are in the inspector. If a button is on top of another, only the top button should be hovered.
There is also a new "ButtonScript", used to test the button eventDispatcher. If you add this script to a gameObject with the buttonComponent, in the Init function the script adds a Delegate to the button, so when it's clicked a function from the ButtonScript is called. Right now this function is just a GLOG, but it is where the button functionality would go. The scripting only works when in play mode.

To test: as with other UI tasks, create buttons and try to use them, along with other UI elements. Load and save to see if everything is fine.


SobrassadaEngine/Scene/Components/Standalone/UI/CanvasComponent.cpp
```diff
@@ -195,5 +180,67 @@ void CanvasComponent::OnWindowResize(const unsigned int width, const unsigned in
...
+        // Only render UI components
+        Component* transform2D = child->GetComponentByType(COMPONENT_TRANSFORM_2D);
+        if (transform2D) static_cast<Transform2DComponent*>(transform2D)->AdaptToParentChanges();
+    }
+}
+
+void CanvasComponent::UpdateChildren()
+{
+    // Update the children list every frame in case they are reorderen in hierarchy. In the future to be more optimal
+    // maybe this can be called only when a gameObject is dragged around the hierarchy
```
srLlobet, line 196: Please wrire TODO here

SobrassadaEngine/Scene/Components/Standalone/UI/CanvasComponent.cpp
```diff
@@ -195,5 +180,67 @@ void CanvasComponent::OnWindowResize(const unsigned int width, const unsigned in
...
+                hoveredButton = currentButton;
+                buttonFound   = true;
+            }
+        }
+    }
+}
+
+void CanvasComponent::OnMouseButtonPressed()
+{
+    if (hoveredButton) hoveredButton->OnClick();
```
srLlobet, line 245: check if the button is interactable here or on the OnClick()

srLlobet (changes requested):
Very good job! Some tiny changes, but overall, great!

LukasGanter (changes requested):
Everything works, just the small changes from Joan remain.
How much work would it be to add a pressed texture / color to the button to indicate if the button is pressed?

srLlobet (approved):
Allright! Looking good!

PolCurto:
Fixed the rotation problem.
Regarding the canvas world view, I did that on purpose because the button hovering is calculated with the screen space position, so the canvas hover and click is not fully accurate. If we wanted to have buttons in the 3d environment (which I don't think we will), it then should be calculated using raycasts and AABBs or something like that.

LukasGanter (approved):
Everything works now, good job!

PR #112, opened 2025-04-14
Image based lighting THO-347

caspi99:
You can see the spheres (its the .zip added) and compare with this photo from the slides. The cubemap is different and ours are LDR so some spheres does'nt seem that good. We need to do a task for hdr cubemaps.

image

[glTF.zip](https://github.com/user-attachments/files/19742544/glTF.zip)


IDarkgenesis (approved):
It works, good job!

🏙️

srLlobet (approved):
Those balls lookin' very sexy my guy. Great job! Solve the merge conflicts and merge!

PR #113, opened 2025-04-14
THO-364 QoL Improvements artist feedback

IDarkgenesis:
Adds hability to copy and delete game objects even inside a multiselection.

Also now nodes in Models have their node name instead of the filename (reinport them before testing)

If you want to delete in multiselection use the hotkey, if you try to delete with the button the multiselection  is cleared to avoid further bugs with other options in the engine.

### DIFF
--- a/SobrassadaEngine/Scene/Components/Standalone/AnimationComponent.h
+++ b/SobrassadaEngine/Scene/Components/Standalone/AnimationComponent.h
@@ -34,12 +34,11 @@ class AnimationComponent : public Component
     UID GetAnimationResource() const { GLOG("Resource AnimUID  is: %d", resource) return resource; }
     ResourceAnimation* GetCurrentAnimation() const { return currentAnimResource; }
     AnimController* GetAnimationController() { return animController; }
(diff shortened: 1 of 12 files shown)


SobrassadaEngine/Scene/Components/Standalone/AnimationComponent.h
```diff
@@ -38,8 +38,8 @@ class AnimationComponent : public Component
 
     void SetAnimationResource(UID animResource);
 
-  private:
     void SetBoneMapping();
+  private:
```
srLlobet, line 42: remove redundant private

SobrassadaEngine/Scene/Components/Standalone/AnimationComponent.h
```diff
@@ -38,8 +38,8 @@ class AnimationComponent : public Component
```
srLlobet, line 38: (It doesn't belong to this task but, shouldn't the bone map getter be const std:: <>& ?

srLlobet (changes requested):
This is amazing work! However, multi-selecting gameobjects with mesh components and deleting them activates the focus, instead of deleting them, and if you then try to multi-select them again, the engine crashes.

PR #115, opened 2025-04-15
Task/lg load gltf default material THO-380 THO-381

LukasGanter:
This pr contains necessary adaptions to correctly import artist gltfs
Main changes:

Fixed issue when loading gltfs without material (Default materials diffuse texture is not changed anymore to the fallback texture)
GLTF loading is now based on nodes rather than meshes. The nodes are iterated and load their assigned meshes and primitives. That solves an issue we had when loading default transforms (It needed the name of the node and mesh to be the same which was not the case for artist gltfs)

For testing: Import gltfs and see if everything works as previously


IDarkgenesis (approved):
Works Good job!
:shipit:

PR #116, opened 2025-04-15
Merge for release version 0.3.3

LukasGanter:
Please test the whole engine, create bug reports for bugs which are not critical and request changes only for issues which cause the engine to crash or make it unusable


srLlobet:
Not a major issue, but if it's easy and quick, being able to delete prefabs from the list would be amazing for level design

srLlobet (approved):
I tried to break the engine for a while, but appart from a weird edge case I reported as a bug, the only major thing I found is that prefabs don't have animations. But that can be fixed with the next release.

IDarkgenesis (changes requested):
Wait for a Quick fix PR for multiselect, there is one case when it crashes when duplicating a multiselect that contains some colliders

IDarkgenesis (approved):
xd

PR #119, opened 2025-04-16
Final State Machine -THO-331

PabioGF:
The final iteration of the state machine is finally here!

Now you can attach an animation to each state of the state machine. To do this, open the state inspector and select a loaded animation.
You can also add triggers and reference them in the transitions, as well as, if the trigger button is pressed, it will change from one state to another.
In the editor the currently active state is shown in green.

You can also add a state machine to the animation component, so you can run it without having to open the editor.

And yes, the interpolation between animations works!

To play an animation on the state machine, load the state machine on the editor or load it on the animation component. Then press Play on the animation component and see the magic. Then click on the triggers to change between states.

Try to run different states with their animations. I leave here attached the zip of the Zombunny with two animations, which are the same but one with less bones, and a fox that has three animations between which you can interpolate.

Please, if the change you are going to request is not critical, accept the PR because my engine doesen't work since the last VS2 instance and i can't work in anything until is solved.

There is still the visual bug that it doesn't draw two inputs in the same node. I know what's the problem, is something internal of ImNodeFlow, but i don't have time to fix it for VS2

https://github.com/user-attachments/assets/73c89f25-96d8-4674-abba-d5cc0b14249d

[Fox (1).zip](https://github.com/user-attachments/files/19782452/Fox.1.zip)

[ZomBunnySpecular (2).zip](https://github.com/user-attachments/files/19782454/ZomBunnySpecular.2.zip)


SobrassadaEngine/UI/EngineEditors/Editor/StateMachineEditor.h
```diff
@@ -25,20 +28,24 @@ class StateMachineEditor : public EngineEditorBase
     void RemoveStateNode(StateNode& node);
     void ShowInspector();
     void DeleteStateResource(StateNode& node);
+    void ShowTriggers();
```
SirKaza, line 31: const

SobrassadaEngine/UI/EngineEditors/Editor/StateMachineEditor.h
```diff
@@ -25,20 +28,24 @@ class StateMachineEditor : public EngineEditorBase
...
+    void ShowTriggersPopup();
```
SirKaza, line 32: const

SobrassadaEngine/UI/EngineEditors/Editor/StateMachineEditor.h
```diff
@@ -25,20 +28,24 @@ class StateMachineEditor : public EngineEditorBase
...
 
 
   private:
     UID uid;
```
SirKaza, line 36: Needed two UIDs? There is one in EngineEditorBase with inheritance

SobrassadaEngine/Scene/Components/Standalone/AnimController.h
```diff
@@ -1,7 +1,10 @@
 #pragma once
+#include "AnimationComponent.h"
 #include "Globals.h"
 
 class ResourceAnimation;
+class Channel;
```
SirKaza, line 6: struct instead of class. This fixes warnings building.
In ResourceAnimation is defined as a struct.

SobrassadaEngine/Scene/Components/Standalone/AnimController.cpp
```diff
@@ -74,27 +75,85 @@ update_status AnimController::Update(float deltaTime)
...
+    {
+        Channel* animChannel = currentAnimation->GetChannel(nodeName);
+        if (animChannel == nullptr) return;
+
+        GetChannelPosition(animChannel, pos, currentTime);
+        GetChannelRotation(animChannel, rot, currentTime);
+    } else
+    {
+        float weight = transitionTime != 0 ? fadeTime / transitionTime : 1;
+        GLOG("transitionTime: %f, fadeTime: %f, weight: %f", transitionTime, fadeTime, weight);
```
SirKaza, line 124: comment this before merge

SirKaza (approved):
Nicely done! Everything works smoothly!
:shipit:

MarcLopezS (approved):
Everything works as expected. Good job! 😄

PR #123, opened 2025-04-18
Task/Audio source component THO-352 & Task/Audio listener component THO-353

PolCurto:
Created the audio source and audio listener components.

Only one audio listener can exist, if a second one is created it does nothing. The audio listener does nothing, it just sends its position and orientation to wwise to be used afterwards. With no audio listener you hear nothing.

The audio source has all the sound data, and also sends its position to wwise to emit sounds. In the editor the user can modify the default event, the volume, the pitch and the 3d spatialization. For now, you can choose a default event (The list of events is parsed from a json generated by wwise) and press the '0' key to play the sounds.

IMPORTANT: I've uploaded the wwise project and the generated soundbanks to a branch called Wwise-project in the game repository. In order to test this PR, you should load the project there so the soundbanks exist. You don't need to download wwise, with the soundbanks that are generated is enough for it to work, but if you want to play around with sounds then you need the wwise sdk.
If you play around with Wwise be careful where files are generated. The soundbanks go in the project folder Assets/Soundbanks/Windows. Generating soundbanks also generates a header file which contains all the IDs that wwise uses, and this file goes to the Utils folder of the engine. You can also use strings with the wwise library, but IDs are faster.

To modify more specific things, like a sound fallof, adding new game parameters, setting states, switches, etc. All that will be done through wwise, and then implemented in the corresponding scripts that need it (player, enemy, boss, etc.). I've created some helper functions in the audio source component to modify wwise game parameters and switches in the future, but right now they are not used. Also, an audio souce can emit any event through scripts, the default one is only for comfort if a source will always uses the same.


SobrassadaEngine/Modules/AudioModule.h
```diff
@@ -1,8 +1,11 @@
 #pragma once
 
 #include "Module.h"
+#include "Standalone/Audio/AudioListenerComponent.h"
+#include "Standalone/Audio/AudioSourceComponent.h"
 
 #include <AkFilePackageLowLevelIODeferred.h>
```
IDarkgenesis, line 7: Please use "" for includes from the Lib folder, and leave <> for standard or libs outside the project eg Windows.h

SobrassadaEngine/Modules/AudioModule.h
```diff
@@ -1,8 +1,11 @@
 #pragma once
 
 #include "Module.h"
+#include "Standalone/Audio/AudioListenerComponent.h"
```
IDarkgenesis, line 4: Can be forward declared

SobrassadaEngine/Modules/AudioModule.h
```diff
@@ -1,8 +1,11 @@
...
+#include "Standalone/Audio/AudioSourceComponent.h"
```
IDarkgenesis, line 5: Can be forward declared

SobrassadaEngine/Modules/AudioModule.h
```diff
@@ -1,8 +1,11 @@
...
 
 #include <AkFilePackageLowLevelIODeferred.h>
+#include <vector>
```
IDarkgenesis, line 9: Add the include for <unordered_map>

SobrassadaEngine/Modules/AudioModule.h
```diff
@@ -14,14 +17,32 @@ class AudioModule : public Module
...
     int music = 0;
     int sound = 0;
     int voice = 0;
     int g_envMAP[255];
+
+    bool loadedAudio = false;
+
+    std::vector<AudioSourceComponent*> sources;
+    AudioListenerComponent* listener;
+    std::unordered_map<std::string, uint32_t> eventsMap;
```
IDarkgenesis, line 47: Maybe not now but use string hashes instead of string for maps, I have seen them being used somewhere in the engine

SobrassadaEngine/Scene/Components/Standalone/Audio/AudioListenerComponent.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include "Application.h"
```
IDarkgenesis, line 3: Remove this include, already in cpp

SobrassadaEngine/Scene/Components/Standalone/Audio/AudioSourceComponent.cpp
```diff
@@ -0,0 +1,174 @@
+#include "AudioSourceComponent.h"
+
+#include "Application.h"
+#include "AudioModule.h"
+#include "EditorUIModule.h"
+#include "GameObject.h"
+#include "InputModule.h" // TODO:  Delete this after testing
+
+#include "ImGui.h"
+#include <AK/SoundEngine/Common/AkSoundEngine.h>
```
IDarkgenesis, line 10: Please use "" for includes from the Lib folder, and leave <> for standard or libs outside the project eg Windows.h

SobrassadaEngine/Scene/Components/Standalone/Audio/AudioSourceComponent.h
```diff
@@ -0,0 +1,46 @@
+#pragma once
+
+#include "Application.h"
```
IDarkgenesis, line 3: Remove this include, already in cpp

SobrassadaEngine/Modules/AudioModule.h
PolCurto: I've added the HashString class in utils so it can be used in other places (Right now it was inside the resourceStateMachine class), but I've left this as a string because the EditorUIModule function to render the list of events needs a string, so maybe that should also be changed in the future

IDarkgenesis (approved):
Thanks for the changes, very good job!

:shipit:

DHazardFenrir (approved):
Great job, tested as requested and I played with the sound distance, pitch and volume, and everything works really well!

PR #126, opened 2025-04-20
THO-348 Deferred shading

IDarkgenesis:
Deferred shading complete. Also fixed and error that crashed the engine / game if it was loaded directly in playmode.

To test: load a scene, if it renders, it works. Also try the game release just in case.


srLlobet (approved):
Very good job!!! Works fine and code is sleek!

PR #127, opened 2025-04-20
Bug/prefabs animations THO-385

PolCurto:
Fixed the cloning animations bug, along with some other small things.

- The AnimationComponent requested the resource each frame, and only when the inspector was shown. That's why it didn't clone animations properly either in prefabs or with Ctrl + D if the inspector was not opened previously. Now it's only called in the Init() and Clone() functions.
- Now cloning an animation also clones the linked resourceStateMachine, if it has.
- Also now you can unlink a gameObject from a prefab, so it is no longer overriden (right click in a prefab in the scene and select "Unlik prefab").

Test if those changes work as expected.

### DIFF
--- a/SobrassadaEngine/Application.h
+++ b/SobrassadaEngine/Application.h
@@ -26,7 +26,7 @@ class AudioModule;
 class EngineTimer;
 class GameTimer;
 
-class Application
+class SOBRASADA_API_ENGINE Application
 {
   public:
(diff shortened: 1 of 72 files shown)


IDarkgenesis (approved):
Nice fixes !

caspi99 (approved):
Good job!

PR #128, opened 2025-04-20
 Bug/RefactoringAnimations/Models broken THO-371

DHazardFenrir:
Changed how animations are imported and added scale to the transforms that should be considered.
The models were broken (the main character and some more) because they had scales, and there were problems when the stride was being calculated. Right now, you can import tthe models with individual animation  in animation test folder(idle, run) should run, you also can test this with the cesiumMan model from Khronos. 

To test:
Import a model (that is not fox or zombunny, preferably the main characters with mixamo anims)
Load that model
Press Play.


PabioGF (approved):
Everithing works perfect. Great job!

MarcLopezS (approved):
I've tested it with the models you gave us and everything works smoothly. I don't see anything weird. Great job! 😄

PR #129, opened 2025-04-20
THO-392 save/load scripting

caspi99:
As I added ImGui, we need to add specific saving and loading from scripts. Also I cleaned the console.
Testing: Just add RotateGameObject Script, change the values and save/load the scene

### DIFF
--- a/SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/Script.cpp
+++ b/SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/Script.cpp
@@ -1,10 +1,11 @@
 #include "pch.h"
-#include "Script.h"
+
 #include "Application.h"
 #include "EditorUIModule.h"
 #include "Math/float2.h"
 #include "Math/float3.h"
 #include "Math/float4.h"
+#include "Script.h"
 
 void Script::Inspector()
 {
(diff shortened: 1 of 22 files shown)


SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/Script.h
```diff
@@ -45,7 +46,9 @@ class Script
 
     virtual bool Init()                  = 0;
     virtual void Update(float deltaTime) = 0;
-    virtual void Inspector()             = 0;
+    virtual void Inspector();
```
SirKaza, line 49: Why virtual these 3 functions?

SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/Script.cpp
```diff
@@ -0,0 +1,123 @@
+#include "pch.h"
```
SirKaza, line 1: separation between these and below so when formatting it works

SobrassadaEngine/Utils/LightsConfig.cpp
```diff
@@ -159,16 +159,15 @@ void LightsConfig::LoadSkyboxTexture(UID resource)
...
 
-        // Im not changing cubemaps right now
-        cubemapIrradiance = CubeMapToTexture(1024, 1024);
+        cubemapIrradiance = CubeMapToTexture(512, 512);
         irradianceHandle  = glGetTextureHandleARB(cubemapIrradiance);
-        glMakeTextureHandleResidentARB(skyboxHandle);
+        glMakeTextureHandleResidentARB(irradianceHandle);
 
-        prefilteredEnvironmentMap       = PreFilteredEnvironmentMapGeneration(1024, 1024);
+        prefilteredEnvironmentMap       = PreFilteredEnvironmentMapGeneration(512, 512);
```
SirKaza, line 166: This 512 somewhere in the header so can be changed easily?

SobrassadaEngine/Utils/LightsConfig.cpp
```diff
@@ -159,16 +159,15 @@ void LightsConfig::LoadSkyboxTexture(UID resource)
...
         prefilteredEnvironmentMapHandle = glGetTextureHandleARB(prefilteredEnvironmentMap);
         glMakeTextureHandleResidentARB(prefilteredEnvironmentMapHandle);
 
-        environmentBRDF       = EnvironmentBRDFGeneration(1024, 1024);
+        environmentBRDF       = EnvironmentBRDFGeneration(512, 512);
```
SirKaza, line 170: same

SirKaza (approved):
Good job with the console, more clear now.
Also with the scripting, so it's set by the scene and not by the scripts.
Check the comments below and tell me what you think.
:shipit:

SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/Script.h
caspi99: If you don't put virtual, the scripts doesnt call the parent function directly. Its does not need to be overwritten by its childs.

IDarkgenesis (approved):
🔢

PR #131, opened 2025-04-21
Player scripting First PR

PabioGF:
A first PR for having the player movement in the VS2.

For testing create a scene and add a Plane. create the navmesh and add a mesh, it can be the fox. Add the player controller, capsule collider and scrpiting components. In the script select the cuchulain. Asign also a cube collider to a child of the fox that will be the weapon. Both Kinematic and in the Player layer and the second fit to size and not generate callbacks.

Create and state machine for the fox with an idle, run, and attack states. Create the triggers (is important to call the triggers the same way you call them in the script(idle,run,attack) change either of both. 

Assign the state machine to the animation component.

When all this is done press the play button and move the fox in the scenary.


IDarkgenesis (approved):
So far, looks like everyting is working

PR #134, opened 2025-04-21
Bug/import existing gltf THO-390

LukasGanter:
Changes the behaviour when reimporting a gltf. Metafiles with the same will be overridden while keeping their UID. The goal is that meshes can be updated later on with other default materials, offsets, normals and so on.
Changes:
- Material and mesh loading order in the mesh component
- Saving of mesh components when using the default material of a mesh (Material will be saved as INVALID_UID)
- Overwriting behaviour when mesh meta files exist already
- Fixed bug which caused prefabs to load every mesh into a separate batch 

For testing:
Use the attached files. Import one, create a scene and add the mesh. Then import the other gltf. After reloading the scene, the material and offset should be updated. 

[TestCube.zip](https://github.com/user-attachments/files/19840903/TestCube.zip)

Previously imported meshes won´t change their materials when they are reimported due to file naming changes and because they have their material saved in the scene file. The game repository is updated to work with these changes.


IDarkgenesis (approved):
Works as expected, also pushed few change to fix a crash in play mode if no camera is present on the scene

srLlobet (commented):
I'll just comment on this one since maybe it's because of my laptop, I'll get back to this when I'm on my pc. However, for me, loading a scene with these meshes displaces all of them and their pivots to the top-right. Moreover, the navmesh doesn't load properly when changing scene. Also, the first time I import the textured mesh and place it, it also appears displaced to the top right, even though it says 0 0 0 transform.

LukasGanter:
> I'll just comment on this one since maybe it's because of my laptop, I'll get back to this when I'm on my pc. However, for me, loading a scene with these meshes displaces all of them and their pivots to the top-right. Moreover, the navmesh doesn't load properly when changing scene. Also, the first time I import the textured mesh and place it, it also appears displaced to the top right, even though it says 0 0 0 transform.

Regarding the navmesh, i added code to prevent a crash on previous scene versions when the navmesh parameter is not present in the scene files. Basically, if its not present, the navmesh uid stays at INVALID_UID. Maybe thats the issue?

Regarding the transform, the textured mesh has a default displacement of the pivot which is not visible in the transform window

LukasGanter:
> Captura These were stacked up when I saved the scene

I don´t know what you mean by that

srLlobet (changes requested):
Yep, everything works properly, except, if there IS a navmesh field on the scene, it gets deleted when saving, so the navmesh never saves regardless of the scene version.

srLlobet (approved):
Nevermind I'm dumb as fuck it works fine. sorry for the scare xD

PR #138, opened 2025-04-26
bug/main char without aabb THO-365

SirKaza:
Trying to fix AABB from artist models I found that Init of Game Objects was not doing it by hierarchy, so this could issue to problems in the future, also fixed some warnings.
Moreover, deleted some game files that should not be in engine.


SobrassadaEngine/Scene/GameObjects/GameObject.cpp
```diff
@@ -477,9 +506,15 @@ AnimationComponent* GameObject::GetAnimationComponent() const
...
 void GameObject::OnTransformUpdated()
 {
-    globalTransform              = GetParentGlobalTransform() * localTransform;
-    globalOBB                    = globalTransform * OBB(localAABB);
-    globalAABB                   = AABB(globalOBB);
+    globalTransform = GetParentGlobalTransform() * localTransform;
+    globalOBB       = globalTransform * OBB(localAABB);
+    globalAABB      = AABB(globalOBB);
+
+    if (globalAABB.IsFinite())
```
IDarkgenesis, line 513: Debug code, maybe?

IDarkgenesis (changes requested):
Nice fix, just check if the mentioned code is necessary

SobrassadaEngine/Scene/GameObjects/GameObject.cpp
SirKaza: Yeah, I will remove it

SobrassadaEngine/Scene/GameObjects/GameObject.cpp
```diff
@@ -149,6 +149,7 @@ GameObject::~GameObject()
 
 void GameObject::Init()
 {
+    if (name == "CH_MC_Chu_V02") int x = 0;
```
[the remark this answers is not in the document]
SirKaza: fixed

IDarkgenesis (approved):
🍌

PolCurto (approved):
Properly generates cuchulan aabb 👍

PR #139, opened 2025-04-26
THO-398 NavmeshCheckbox and New components storage

IDarkgenesis:
This PR changes the navmesh creation to only use the game objects that have the navmesh checkbox activated, important, in order to take effect the checkbox must be in the GO that contains the mesh. 

Also changes the storage of Components to a tuple so dynamics casts can be avoided everywhere components are needed. There is a templated GetComponent now, just ask for the specific component*.

Also fixes a crash for not checking if script components inside the onCollision functions weren't nullptr.


PolCurto (approved):
The navmesh is properly applied only to the checked meshes, Regarding the components I've played around with most of them and everything seems fine, there was an issue with the button I already fixed. Good job! 🍶

caspi99 (approved):
Seems all good! 👯‍♀️ 🫓

PR #142, opened 2025-05-01
THO-464 Prefab fixing

IDarkgenesis:
Now updating a prefab overwrites the existing one. Also added capability to delete prefabs from the load prefab menu.

Also fixed an issue that made prefabs spawn at 000 when loading a scene

### DIFF
--- a/SobrassadaEngine/FileSystem/PrefabManager.cpp
+++ b/SobrassadaEngine/FileSystem/PrefabManager.cpp
@@ -32,25 +32,7 @@ namespace PrefabManager
         const std::string& savePath = App->GetProjectModule()->GetLoadedProjectPath() + PREFABS_LIB_PATH +
                                       std::to_string(finalPrefabUID) + PREFAB_EXTENSION;
 
-        const std::string& originalName = gameObject->GetName();
-
-        int numRepeated                 = 0;
-        bool updateName                 = true;
-        std::string name                = std::string(originalName);
-        while (updateName)
-        {
-            updateName = false;
-            for (const auto& savedPrefab : App->GetLibraryModule()->GetPrefabMap())
-            {
-                if (savedPrefab.first == name)
-                {
-                    updateName = true;
-                    ++numRepeated;
-                    name = originalName + "(" + std::to_string(numRepeated) + ")";
-                    break;
-                }
-            }
-        }
+        std::string name = gameObject->GetName();
 
         // Create structure
         prefab.AddMember("UID", finalPrefabUID, allocator);
--- a/SobrassadaEngine/Scene/Components/Standalone/MeshComponent.cpp
+++ b/SobrassadaEngine/Scene/Components/Standalone/MeshComponent.cpp
@@ -108,8 +108,11 @@ void MeshComponent::Clone(const Component* other)
         const MeshComponent* otherMesh = static_cast<const MeshComponent*>(other);
         enabled                        = otherMesh->enabled;
 
-        AddMesh(otherMesh->currentMesh->GetUID());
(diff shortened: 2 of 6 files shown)


SobrassadaEngine/FileSystem/PrefabManager.cpp
```diff
@@ -32,25 +32,7 @@ namespace PrefabManager
...
-                if (savedPrefab.first == name)
-                {
-                    updateName = true;
-                    ++numRepeated;
-                    name = originalName + "(" + std::to_string(numRepeated) + ")";
-                    break;
-                }
-            }
-        }
+        std::string name = gameObject->GetName();
```
SirKaza, line 35: const &

SobrassadaEngine/Scene/Components/Standalone/MeshComponent.cpp
```diff
@@ -108,8 +108,11 @@ void MeshComponent::Clone(const Component* other)
         const MeshComponent* otherMesh = static_cast<const MeshComponent*>(other);
         enabled                        = otherMesh->enabled;
 
-        AddMesh(otherMesh->currentMesh->GetUID());
-        AddMaterial(otherMesh->currentMaterial->GetUID());
+        UID otherMeshUID               = otherMesh->currentMesh ? otherMesh->currentMesh->GetUID() : INVALID_UID;
```
SirKaza, line 111: const

SobrassadaEngine/Scene/Components/Standalone/MeshComponent.cpp
```diff
@@ -108,8 +108,11 @@ void MeshComponent::Clone(const Component* other)
...
+        UID otherMatUID = otherMesh->currentMaterial ? otherMesh->currentMaterial->GetUID() : INVALID_UID;
```
SirKaza, line 112: const

LukasGanter (changes requested):
I have to test a bit more, but i found issues with deleting and updating

LukasGanter (changes requested):
In the same project, updating a duplicated prefab (for example "Prefab_Island_2_Lvl_1(3)" duplicates the file once and creates "Prefab_Island_2_Lvl_1".
Solution here can be either that we update all duplicated prefabs once manually and remove the (x) tags, or to add these again to the name when updating.

IDarkgenesis:
> Deleting previously duplicated prefabs doesn´t work. Try deleting the prefab "Prefab_CityCenter(1)" in the game repository, branch verticalSlice/VS3

Due to the old system Prefab_CityCenter(1) and Prefab_CityCenter share the same UID and the last one to be loaded into the library is Prefab_CityCenter, which overrides the name in the map so it only deletes Prefab_CityCenter. Needs to be deleted manually, in the new version this does not happen since overrides the prefab.

IDarkgenesis:
> In the same project, updating a duplicated prefab (for example "Prefab_Island_2_Lvl_1(3)" duplicates the file once and creates "Prefab_Island_2_Lvl_1". Solution here can be either that we update all duplicated prefabs once manually and remove the (x) tags, or to add these again to the name when updating.

Similar to the previous comment this is an issue of the old save system, since they have the same UID only the Prefab_Island_2_Lvl_1 will be treated as the real prefab all others must be deleted by hand, from now on it gets overwritten

IDarkgenesis:
Solution could be to update the current prefabs to genertate the one with the version without brackets and the remove the old ones, from that point on, we shouldn't have this issues

LukasGanter (approved):
Issues will be adressed manually in the game repository

PR #143, opened 2025-05-01
THO-459 multiple script per game object

caspi99:
Testing:
You can add multiple scripts now. In rotateGameObject there are an imgui gameobject option now.
Also I solved part of the leaks of the play/stop button (not all of them) but now there's only 100mb of memory leak instead of 300mb


IDarkgenesis (approved):
Worked without issues, nice addition the selection for GameObjects as target for scripts.

👍

PolCurto (approved):
Works as expected, good job! Keep in mind that if you select a prefab as the gameObject target, each time the scene is loaded the link will be broken 🦺

PR #144, opened 2025-05-02
THO-476 Invert mouse picking

IDarkgenesis:
Mouse picking inverted, so now only nodes with select parent checkbox skip to the next parent, when loading a model, all children are set with this checkbox.

Also, now clicking on a game object expands the tree view upwards.

### DIFF
--- a/SobrassadaEngine/Scene/Scene.cpp
+++ b/SobrassadaEngine/Scene/Scene.cpp
@@ -1111,6 +1111,8 @@ void Scene::LoadModel(const UID modelUID)
                                 currentGameObject->GetName() + " Mesh " + std::to_string(meshNum)
                             );
                             ++meshNum;
+
+                            gameObjectsArray.push_back(meshObject);
                         }
                         else
                         {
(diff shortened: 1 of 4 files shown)


SobrassadaEngine/Scene/Scene.cpp
LukasGanter, line 1: Line 1189, the index starts as 1 which leads to the first game object not being marked

LukasGanter (changes requested):
Apart from the index, everything works as expected.
Great work!

SobrassadaEngine/Scene/Scene.cpp
IDarkgenesis: This is done on purpouse because the 0 element is the parent of the model, if we select it as true, it will select the scene root game object

caspi99 (approved):
🗼 ✈️ 💥

SobrassadaEngine/Scene/Scene.cpp
LukasGanter: Did you implement the fix that we talked about already?

IDarkgenesis: Should be in now and working

LukasGanter: Works now

LukasGanter (approved):
Works now as expected, good job!

PR #145, opened 2025-05-02
THO-460 Multi use state machine

IDarkgenesis:
Now state machines are still a resource but the active state has been moved to the animation component allowing multiple components to use the same state machine without affecting others.

SharedStateMachine

If you don't have a model with different animations you can use this one, to build and test a state machine.
[Fox.zip](https://github.com/user-attachments/files/20015143/Fox.zip)

### DIFF
--- a/SobrassadaEngine/FileSystem/StateMachineManager.cpp
+++ b/SobrassadaEngine/FileSystem/StateMachineManager.cpp
@@ -229,7 +229,6 @@ namespace StateMachineManager
         }
 
         stateMachine->SetDefaultState(0);
-        stateMachine->SetActiveState(0);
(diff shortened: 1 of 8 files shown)


PabioGF (changes requested):
With the state machine associated in the component. When clicking stop with the fox running and then clicking Play again the active state should be the default state, in the fox case idle, not running.  Also, now that the active state is not in the resource, is not drawn in the editor the current active state, is this a problem?

PabioGF (approved):
Works perfect, great job!

SirKaza (approved):
Tested in debug and release.
Works perfectly. Animations are still a problem when cloning to another project.

PR #146, opened 2025-05-02
Task/Player Camera & Movement Rework THO-408

PolCurto:
Updated camera and player movement.

There is a scene in the EngineTests repo, in the branch CameraTesting, where you can test this.
The camera has some parameters you can change: smoothness, mouse offset, lookAhead, follow distance threshold, etc. Tweak them and see what works better. 
With the left mouse click, the player attacks, and with the right click you can aim. While it does this it stops moving. You can change its speed and acceleration.


LukasGanter (changes requested):
Is it possible to combine the player mesh and camera pivot into one prefab? So that level designers just need to include one prefab and are ready to go?

PolCurto:
Yes, you can put the player and the camera as children of the same gameObject and create a prefab of it.

PolCurto:
There is an issue that prefabs didn't store the values in scripts, which is fixed in the next PR of the Ranged Attack, so it would be nice to merge this one asap

PR #147, opened 2025-05-04
THO-485 camera in prefab not working

caspi99:
You can delete cameraComponent now. Also camera in prefabs works now. Also solved a bug where skybox does not work in a new scene
For testing: You can try all of these things


PolCurto (approved):
🩻

srLlobet (approved):
Lookin' good

PR #149, opened 2025-05-04
task/Free Camera THO-477

SirKaza:
Simple FreeCamera script and prefab for navigating through the level while we don't have player.

How to test:
Create an object and add camera and script components (freecamera script).
Set it as a prefab and test that it works.


MarcLopezS (approved):
Tested and everything works perfectly. Good Job! 😄

LukasGanter (approved):
Works great, good job

PR #150, opened 2025-05-05
Meta override while keeping UID

LukasGanter:
All metafiles now keep their UID if they existed before and are overwritten by a new import.
This ensures that previous established links, for example between a mesh and its material, are not broken after a reimport.
Test with importing various files and see if the meta files keep their UID

### DIFF
--- a/SobrassadaEngine/FileSystem/Importers/AnimationImporter.cpp
+++ b/SobrassadaEngine/FileSystem/Importers/AnimationImporter.cpp
@@ -210,9 +210,15 @@ namespace AnimationImporter
         UID finalAnimUID;
         if (sourceUID == INVALID_UID)
         {
-            const UID animationUID      = GenerateUID();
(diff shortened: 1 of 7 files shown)


IDarkgenesis (approved):
Have not found any issues

SirKaza (changes requested):
Reimporting a gltf that is inside assets deletes the gltf and the binary.
Furthermore, models don't work and if library is deleted and assets/gltf aren't it crashes in ImportMesh from MeshImporter.
If reimport is from outside this folder it works correctly, I suppose this is not expected.

IDarkgenesis (changes requested):
There is an issue when calculating the name of the meta when checking UID, at least with animations the name when importing is not the same as the check leading to reimporting the animations again with a different UID.

image

IDarkgenesis (approved):
Works

SirKaza (approved):

PR #155, opened 2025-05-06
Task/Respawn-THO-487

PolCurto:
Added spawn points and respawn on key press for easier game navigation and debugging. Also fixed a bug that the characterController was going through the floor when not in y = 0 and being child of another object.

Pressing F6 saves the current player position as the spawn point, and pressing F5 spawns the player in the saved position. The default is 0,0,0. 

There is a new script SpawnPoint, which is basically a trigger that sets its position as the player spawn position when it enters the trigger. This can be used as the spawn points in the levels. 
In the EngineTests dev-testing branch there is a new scene with a Spawn prefab that has already the script and collider to work properly. This scene has also terrain in differen heights, to check the camera behaves as expected anywhere.

Known issue: Right now the player projectile collides with triggers, which disables it although it is not really colliding with anything. Probably a new physics layer should be added for projectiles so it doesn't collide with triggers.

### DIFF
--- a/SobrassadaEngine/Scene/Components/ScriptComponent.h
+++ b/SobrassadaEngine/Scene/Components/ScriptComponent.h
@@ -20,6 +20,7 @@ enum ScriptType
     SCRIPT_CAMERA_MOVEMENT,
     SCRIPT_PROJECTILE,
     SCRIPT_FREE_CAMERA,
+    SCRIPT_SPAWN_POINT,
 
     SCRIPT_TYPE_COUNT // Add at the end
 };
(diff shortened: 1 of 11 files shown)


LukasGanter (approved):
Works as expected, good job!
Is it possible to extend the system later to have multiple spawn points and then selecting the last one passed?

LukasGanter (commented):
The new physics layer mentioned as a know issue here will be implemented in a separate task (https://fpcvideogames.youtrack.cloud/issue/THO-496/Projectile-physics-layer)

IDarkgenesis (approved):
Great job, works as expected!

🚗

PR #156, opened 2025-05-06
THO-484 Meta animation wrong file path

IDarkgenesis:
This PR fixes few things with the animation import and component.

Animation are now properly reimported from metafiles if the library is deleted, also selecting an animation in the dropdown of the component editor assigns that animation to the component.

In order for the changes to work existing models with animations need to be reimported so the metafile gets updated with the index in the model.

You can try with the artist model which contains one animation and with the fox which contains 3, once imported and deleting the library, once the scene is opened the models should mantain their animations.

### DIFF
--- a/SobrassadaEngine/Utils/Globals.h
+++ b/SobrassadaEngine/Utils/Globals.h
@@ -95,7 +95,7 @@ constexpr const char* SCENES_PATH                    = "Assets/Scenes/";
 constexpr const char* METADATA_PATH                  = "Assets/Metadata/";
 constexpr const char* PREFABS_ASSETS_PATH            = "Assets/Prefabs/";
 constexpr const char* MODELS_ASSETS_PATH             = "Assets/Models/";
-constexpr const char* ANIMATIONS_ASSETS_PATH         = "Assets/Animations/";
+constexpr const char* ANIMATIONS_ASSETS_PATH         = "Assets/";
 constexpr const char* STATEMACHINES_ASSETS_PATH      = "Assets/StateMachines/";
 
 constexpr const char* LIBRARY_PATH                   = "Library/";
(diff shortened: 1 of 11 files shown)


SobrassadaEngine/Utils/Globals.h
```diff
@@ -95,7 +95,7 @@ constexpr const char* SCENES_PATH                    = "Assets/Scenes/";
 constexpr const char* METADATA_PATH                  = "Assets/Metadata/";
 constexpr const char* PREFABS_ASSETS_PATH            = "Assets/Prefabs/";
 constexpr const char* MODELS_ASSETS_PATH             = "Assets/Models/";
-constexpr const char* ANIMATIONS_ASSETS_PATH         = "Assets/Animations/";
+constexpr const char* ANIMATIONS_ASSETS_PATH         = "Assets/";
```
SirKaza, line 98: Delete this variable and use assets_path to find the gltf files, that is what is intented for

SirKaza (changes requested):
Reimporting a model doesn't keep the animation and has to be applied again. Don't know if this is expected.
Also, it's only a visual bug but deleting the whole library and then executing the engine, this execution will show double the amount of animation that it really has.
It seems that in some cases the names changed if it's imported first time. Untitled2_Run instead of Run when reimporting also. Might be only a visual bug of Animation Library selection in the editor inspector.

image

SobrassadaEngine/Utils/Globals.h
IDarkgenesis: The amount of animations could be because a change on how names are stored in metas, which will produce new meta names and old ones not being deleted.

Does the animation keep selected after saving the first time, or its always missing, could be for the previous reason.

PolCurto (changes requested):
Overall works good for me, good job. I imported the player and the fox, changed the fox animation and deleted the library. When reopneing the engine, all the animations were there. 
However, if importing the player or the fox again (without deleting the lib or anything) causes this new model to have all animations, at least for me. For example the player has also the fox animations, and vice versa. I don't know if it's related with what Marc said.
image

PolCurto (approved):
🍡

SirKaza (approved):
🍡

PR #158, opened 2025-05-07
THO-496 Added new projectile layers for player and enemy

IDarkgenesis:
Added two new physics layers the player and enemy projectile.

**Important!** If you load an existing scene you will need to add the collision config on the editor settings, if a new scene is created it will take the default values.

image

Default values:
Player projectile -> enemies
Enemy projectile -> player
Player-> also interacts with enemy projectile


PolCurto (approved):
📵

caspi99 (approved):
👻

PR #159, opened 2025-05-07
THO-467 string hashes in maps

caspi99:
I don't know how you can test this. Just execute and see it does not crash. Also solved 2 OpenGL errors.
I can't do the forward declaration for this part of code:
namespace std
{
    template <> struct hash<HashString>
    {
        size_t operator()(const HashString& hs) const { return hash<string> {}(hs.original); }
    };
}


srLlobet (approved):
(☉౪ ⊙)

IDarkgenesis (approved):
👎

PR #160, opened 2025-05-07
THO-486 add new game object

caspi99:
Test: You can do right-click on the hierarchy window and create a new GameObject


SirKaza (changes requested):
Is not working in debug and release.

IDarkgenesis (approved):
yEs'

SirKaza (commented):
If it is possible, would be nice that right click in hierarchy makes children of the scene root game object. 
And when right-clicking a specific game object and adding a new one, it is children of this selected.

SirKaza (approved):
:shipit:

PR #161, opened 2025-05-07
bug/Player falls in game THO-489

SirKaza:
Fixed the bug that in game mode, player was falling through nav mesh.
Also, fixed paused time, so logic is paused when pause menu is opened.
Init of nav mesh only once in the constructor, so enemies have the nav mesh loaded correctly.

How to test:
Have a model with character controller and run game config and see if it works.


caspi99 (approved):
It works ♐

IDarkgenesis (approved):
Si,

SirKaza:
Don't merge yet. Pol has to do a change

PR #166, opened 2025-05-08
Fix aiming in game mode

PolCurto:
FIxes the aiming and attacking direction of the player in game and gameDebug mode.

To test: execute the engine in game or gameDebug and see if it aims and attacks towards where the mouse is

Also fixed that characterController rotation in radians was applied as degrees anyway

### DIFF
--- a/SobrassadaEngine/Scene/Components/CameraComponent.cpp
+++ b/SobrassadaEngine/Scene/Components/CameraComponent.cpp
@@ -8,6 +8,7 @@
 #include "InputModule.h"
 #include "OpenGLModule.h"
 #include "SceneModule.h"
+#include "WindowModule.h"
 
 #include "ImGui.h"
 #include "Math/MathFunc.h"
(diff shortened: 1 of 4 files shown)


SobrassadaEngine/Scene/Components/CameraComponent.cpp
```diff
@@ -423,6 +424,17 @@ void SOBRASADA_API_ENGINE CameraComponent::Rotate(float yaw, float pitch)
 
 const LineSegment CameraComponent::CastCameraRay()
 {
+#ifdef GAME
+    // In game, just get the mouse position because it covers all the window
+    const float2 mousePosition    = App->GetInputModule()->GetMousePosition();
```
SirKaza, line 429: &

SirKaza (approved):
I suppose it works. Its crashing my pc every time I try to run in game cause createScript function:
image

image

I tried cleaning solution deleting dlls and it keeps crashing

caspi99 (approved):
🦖

MarcLopezS:
Works very smoothly. Well done! 🥇

PR #167, opened 2025-05-08
Game-HUD-THO-341

jmartiolivart:
### The main changes are:
- Refactored Transform2DComponent to preserve local position correctly when switching between render modes.
- Fixed rendering offsets by applying parent transform correction only in World Space.
- Added automatic creation of CanvasScalerComponent to enable consistent scaling based on resolution.
- Ensured CanvasComponent handles bounding box updates more reliably.
- Fixed cumulative position errors when toggling between Screen Space and World Space.
- Applied all necessary corrections to UILabelComponent and ImageComponent.
- Verified ButtonComponent behavior remained functional without additional changes.

### How to test:

- Create a UI canvas with buttons or labels.
- Switch render mode between Screen Space Overlay and World Space — UI should keep its position and scale.
- Move elements with anchors and pivot set — positions should remain visually stable.
. Resize the window — UI should scale properly if CanvasScaler is present.
- Ensure buttons still respond to mouse and keyboard input.
- Save/load scenes and confirm transform states persist correctly.


SobrassadaEngine/Scene/Components/Standalone/UI/ButtonComponent.cpp
```diff
@@ -143,6 +143,12 @@ void ButtonComponent::Clone(const Component* other)
 
 void ButtonComponent::RenderDebug(float deltaTime)
 {
+    CanvasComponent* canvas = parent->GetComponentParent<CanvasComponent*>(App);
+    if (canvas != nullptr)
```
PolCurto, line 147: Don't need to get the canvas each frame, ButtonComponent already has a pointer to it as a member

SobrassadaEngine/Scene/Components/Standalone/UI/ImageComponent.cpp
```diff
@@ -114,6 +114,12 @@ void ImageComponent::Clone(const Component* other)
 
 void ImageComponent::RenderDebug(float deltaTime)
 {
+    CanvasComponent* canvas = parent->GetComponentParent<CanvasComponent*>(App);
+    if (canvas != nullptr)
```
PolCurto, line 118: Don't need to get the canvas every frame, ImageComponent already has a pointer to it as a member

PolCurto (changes requested):
Tested the UI with all components, here are some issues I found:
- Changing canvas from world space to screen space makes all components go to 0,0. From screen space to world space works fine
- Buttons don't seem to work fine, specially when having more than one. Only the one upfront seems that can be clicked

https://github.com/user-attachments/assets/a1e3fc84-ade6-4045-8dfc-93edcf4d5b50



Some minor issues:
- Focusing the canvas makes you go so far it can't even be seen because of the far plane. You can just zoom in, but maybe making the aabb smaller helps the focus not go so far
- Before, canvas took into account the scene frameBuffer to resize, so elements were not stretched out, and avoids issues wth button hover in the engine. This, however, is not that important because in game mode the game occupies all the window so it will work fine. It's only for comfortability in the engine.

Also, is the match parent size checkbox needed? Can't that be done with anchors?

PolCurto (approved):

SobrassadaEngine/Scene/Components/Standalone/UI/ButtonComponent.cpp
```diff
@@ -163,8 +167,17 @@ bool ButtonComponent::UpdateMousePosition(const float2& mousePos, bool dismiss)
 {
     if (!isInteractable || !IsEffectivelyEnabled()) return false;
 
+    GLOG(
```
SirKaza, line 170: comment some logs before merging

SobrassadaEngine/Scene/Components/Standalone/UI/Transform2DComponent.h
```diff
@@ -63,8 +67,8 @@ class Transform2DComponent : public Component
     float2 anchorsY;
 
   private:
-    CanvasComponent* parentCanvas;
-    Transform2DComponent* parentTransform;
+    CanvasComponent* parentCanvas         = nullptr;
+    Transform2DComponent* parentTransform = nullptr;
     std::vector<Transform2DComponent*> childTransforms;
 
     bool transform2DUpdated;
```
SirKaza, line 74: init header variables, also there is a static variable at the top with the methods, if possible change it so it's not static and put it here

SirKaza (approved):
I tried to see if something was wrong, but everything works as expected. Fantastic job!

🥖 🥖 
image

LukasGanter (approved):
Build in release mode works

PR #169, opened 2025-05-08
Fixed crash when a broken prefab file is being loaded

LukasGanter:
This PR fixed a bug when the user attempts to load a prefab with a corrupted file. 

For testing, modify a prefab file so that the json syntax is broken. 
The engine should not crash and a log with the error is displayed in the console.


PolCurto (approved):
⚡

IDarkgenesis (approved):
👍

PR #172, opened 2025-05-09
THO-361 project specific scripting

caspi99:
For testing: In dev-testing there is an scene called spawnTesting with a sound. You can check that it does not crash and its good.
Also added the case insensitive for searching components or meshes/materials


PolCurto:
The sound plays when shooting the spear

PR #173, opened 2025-05-10
Allowed textures to be overwritten by new versions THO-502

LukasGanter:
This pr fixes an import problem which prevented textures with the same name to be overwritten by new versions. 

To test, go to the AssetOverview scene of the EngineTest repo and import the following gltf: 
https://drive.google.com/file/d/1xKrhXlu4QHt5E88QBw5jTtoMhnm1l2uw/view?usp=sharing

After importing, the pillars should switch from a blue texture to a grey texture. Before this change they stayed blue.


IDarkgenesis (approved):
Changed the texture

caspi99 (approved):
👍

PR #174, opened 2025-05-10
THO-465 Disabled component / game object 

caspi99:
Testing: I put a branch in engine-Test. Now you can enable/disable a parent and its affects all the components and childs. Also now showing ImGui always (I asked some people and they saw this change good). Tried all components and its disabling the functionality (Audio is done in another branch by Pol, so dont check audio). Also added enabled/disabled on scripts, so you can enable/disable only 1 script.


IDarkgenesis (changes requested):
Very good job, just this small bug with the camera component

cameraDisableBug

MarcLopezS (approved):
Works as expected! Nice! Good job! 😄

IDarkgenesis (approved):
🚒

PR #175, opened 2025-05-10
Task/Enemy Behaviour THO-462

SirKaza:
· Patrol enemy implemented between initial and defined points.
· Improved the soldier chase behavior so it feels more natural.

How to test:
In engineTests repository branch dev-testing there is a scene named enemyTesting with all new enemy features to test it.
To set patrol points, if crowd navigation checkbox is enabled from crowd control window and click to scene points it will print exactly the position to introduce it as a patrol point.

### DIFF
--- a/SobrassadaEngine/Scene/Components/Standalone/AIAgentComponent.cpp
+++ b/SobrassadaEngine/Scene/Components/Standalone/AIAgentComponent.cpp
@@ -52,12 +52,19 @@ AIAgentComponent::~AIAgentComponent()
 void AIAgentComponent::Update(float deltaTime)
 {
     if (!IsEffectivelyEnabled()) return;
-
     if (!App->GetSceneModule()->GetInPlayMode()) return;
+
+    dtCrowd* crowd = App->GetPathfinderModule()->GetCrowd();
+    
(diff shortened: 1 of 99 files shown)


SobrassadaEngine/Scene/Components/Standalone/AIAgentComponent.cpp
```diff
@@ -229,9 +297,11 @@ void AIAgentComponent::LookAtMovement(const float3& moveDir, float deltaTime)
 
     if (fabs(angle) < 0.0001f) return;
 
-    float4x4 rotY  = float4x4::FromEulerXYZ(0.0f, angle, 0.0f);
-    float4x4 local = parent->GetGlobalTransform() * rotY;
+    float4x4 rotY      = float4x4::FromEulerXYZ(0.0f, angle, 0.0f);
+    float4x4 newGlobal = parent->GetGlobalTransform() * rotY;
+
+    float4x4 newlocal  = parent->GetParentGlobalTransform().Transposed() * newGlobal;
```
PolCurto, line 303: some here can be const

PolCurto (approved):
🧦

PR #176, opened 2025-05-11
THO-501 navmesh load on scene change

caspi99:
Test: You can try scenes with navmesh and scenes that does not have a navmesh. I left in engineTest in a branch called navmesh 3 different scenes. 2 With navmesh and 1 without.
Also cleared all the glogs that are not an error


IDarkgenesis (approved):
No issues found, good job!

PolCurto (approved):
GLOG 🪦

PR #179, opened 2025-05-12
Task/moving tiles THO-338

srLlobet:
Anything with this script attached will lower itself to the ground and stay there until the player gets close enough. Then it will rise, and when the player stops being close, the mesh will lower itself again.

To test: There's a new scene in dev-testing in the engine tests repo called TileFloatTesting. Just load it, click play, and walk towards the tiles, you should see them lowering and raising themselves as you get close/far.


PolCurto (approved):
Looks good!

LukasGanter (changes requested):
The basic functionality works, but many parameters should be passed as parameters to the script so we can iterate faster. These include: 
- Movement speed (or time it takes to move from default transform to target transform)
- Default transform offset from game object transform 
- Target transform offset from game object transform 
- Triggerbox (Or at least a parameter which lets you define how far away the player can trigger the element from

We want to use the script for moving other things as well, not just tiles. Thats why it needs to be customizable

PR #181, opened 2025-05-12
Fixed bug which loads meshes in the wrong order

LukasGanter:
Fixed bug which loads meshes in the wrong order. The order of the meshes in models in the engine was determined on the node order which broke links in specific cases. Now the mesh order is the same as it is in the gltf.

When importing the provided gltf, the animation should run perfectly.
[Attack1.zip](https://github.com/user-attachments/files/20163850/Attack1.zip)


PolCurto (approved):
Tried the attack animation and now works perfect, good job!

PR #183, opened 2025-05-12
THO-599 Improvement to collider editor

IDarkgenesis:
Changed collider editor to use drag floats to make it simple to adapt use.

Note: for weapon colliders use capsule collider instead of cube, leave those for triggers


PolCurto (approved):
🗄️

LukasGanter (approved):
Works

PR #184, opened 2025-05-14
Task/Enemy banshee V1 THO-574

PolCurto:
Created the banshee script, which adds the behaviour of the banshee as described iin the gdd. The banshee stands still, when the player gets close starts chasing and the attack is made of two areas: an area in front of it and an area around it. If the player gets very close it flees until its far enough to attack again.
While charging the attack, the banshee slowly rotates towards the player position. This rotation speed and the fleeing speed can be modified in the script inspector.

Also fixed some other things (Pathfinder use game deltaTime when in Play Mode to not move enemies when game is paused, enemies look forward when moving and not to the player, so they walk forward and not to the sides when avoiding an obstacle, etc.).

Also now the hitbox is not enabled during all the attack, only during a biref moment after a delay, to have it prepared when animations come in. Both timings can be modified in the inspectore as well.

There is an scene in engine testing branch dev-testing called EnemiesTest, where there is a banshee. There is also a normal soldier patroling to check it still works fine.

### DIFF
--- a/Game/EngineDefaults/Shader/Fragment/BillboardFragment.glsl
+++ b/Game/EngineDefaults/Shader/Fragment/BillboardFragment.glsl
@@ -1,0 +1,12 @@
+#version 460
+
+layout(binding=0) uniform sampler2D myTexture;
+
+in vec2 uv;
+
+out vec4 fragColor;
+
+void main()
+{
+    fragColor = texture2D(myTexture, vec2(uv.x, 1 - uv.y));
+}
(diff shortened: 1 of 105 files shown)


LukasGanter (changes requested):
Works exactly as expected, great job!
Only reason i don´t approve yet is that we might want to add some more functionalities to the behaviour which are not yet in the GDD.

PolCurto:
Added the new behaviours. The PR description is updated and describes the new changes.

LukasGanter (commented):
For me, the player attack as well as all enemy scripts are not working. Moving the player works. I rebuild everything already and copied the script dll manually. Tested with the EnemiesTest scene

IDarkgenesis (approved):
Banshee's behaviour is as expected, soldier still works as intented good job.

🪕

LukasGanter (approved):
Works now as expected, good job!

PR #185, opened 2025-05-15
Controller-Support-THO-498

jmartiolivart:
To test this PR, connect a game controller (tested with a PS5 DualSense controller) and run the engine. 
Move the analog sticks, press triggers and buttons — logs will be printed to the console showing their state. This verifies that controller input is being correctly read and processed.

I tested with: PlayStation 5 DualSense Controller.


LukasGanter (approved):
Works perfectly, great job!

PR #186, opened 2025-05-15
Bug/conditional-controller-bug-THO-605

jmartiolivart:
Check if it works now without crashing without the controller and check that logs stills are working.


PolCurto (approved):
Tested with no controller connected and everythingg worked fine

LukasGanter (approved):
Looks good, no crash anymore

caspi99 (changes requested):
image
It happens this to me when I close the engine with a controller on

caspi99 (approved):
Pol said to me is solved in his branch, so its working the rest

PR #188, opened 2025-05-16
THO-607 Solving error

caspi99:
These changes should solves the scripting error. The problem was that the .lib was generated but in another place so the new scripts were using the old lib. Now you wont have the problem.

### DIFF
--- a/SobrassadaEngine/Modules/ScriptModule.cpp
+++ b/SobrassadaEngine/Modules/ScriptModule.cpp
@@ -48,6 +48,9 @@ void ScriptModule::LoadDLL()
         GLOG("Failed to load DLL\n");
         return;
     }
+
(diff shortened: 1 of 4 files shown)


PolCurto (approved):
Tested mergging with the dash and the banshee scripts, and this solved the bug, great job 🧙‍♂️

SobrassadaEngine/Modules/ScriptModule.cpp
```diff
@@ -48,6 +48,9 @@ void ScriptModule::LoadDLL()
         GLOG("Failed to load DLL\n");
         return;
     }
+
+    if (!fs::exists("SobrassadaScripts.dll")) fs::copy(dllPath, copyPath, fs::copy_options::overwrite_existing);
```
SirKaza, line 52: Use file system namespace functions

IDarkgenesis (approved):
Bona feina. 👍

PR #189, opened 2025-05-16
Task/Player controller controls THO-591

PolCurto:
Adapted the player actions to a controller. I used as a reference for the controls this image, only with the actions we have now. 
Captura_de_pantalla_2025-05-15_161212

It should still work with keyboard as well, test that too just in case. You can test this in any scene with the player, for example the spawnTesting one


IDarkgenesis (changes requested):
Tested with a XBox controller wired and wireless, movement with left joystick works, looks like the normal attack is working but the shooting (right click mouse) and d-pad movement are not doing the expected inputs

IDarkgenesis (approved):
Now works, nice, mouse playing is superior.

LukasGanter (changes requested):
Shooting doesn´t work with a Dualshock 4 controller (Aiming with L2 is not doing anything)

LukasGanter (approved):
The error was on my side, the pr works perfectly. Good job!

PR #191, opened 2025-05-17
Task/Fireball Trap THO- 500

SirKaza:
Implemented the first trap that consists in a fireball falling into a certain spot causing an area of damage for a certain period of time. This trap activates every x time. In the new script named FireballTrap, there are loads of variables to tweak to put it to your liking. Also, created a model that contains the plane where fireball with impact with the model that the artists gave us.

How to test:
In EngineTests repository branch dev-testing, there is a scene called FireTrapTesting where this feature can be tested properly.


IDarkgenesis (approved):
Walking on top of the trap reduces the health as expected.
Don't forger to merge VS3 now contains the fix for linking errors.

Good job! 💯

caspi99 (approved):
Works as expected! 🥇

PR #192, opened 2025-05-17
THO-61 debug view navmesh meshes prefab

caspi99:
Testing: There are 3 new render debugs. GBuffers, Depth where you see the textures and Navmesh meshes. All the meshes with NavMesh Valid are rendered normally and the ones without the checkbox is rendered as wireframe. Remember to have objects on scene with the option and without it to see anything

### DIFF
--- a/Game/EngineDefaults/Shader/Fragment/DepthFragment.glsl
+++ b/Game/EngineDefaults/Shader/Fragment/DepthFragment.glsl
@@ -1,5 +1,7 @@
+#version 460
+
 uniform sampler2D u_Texture;
-in vec2 TexCoords;
+in vec2 uv0;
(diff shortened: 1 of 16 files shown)


IDarkgenesis (approved):
👍

LukasGanter (changes requested):
Changes to the geometry (Location, Rotation, Scale) are not visually updated while the debug view is active

LukasGanter (changes requested):
It looks like the issue with moving the meshes is fixed.
When loading the level 1 scene from the game repository, though, there are errors in the vertex shader which lead to very low fps: 
grafik
This error occurs with all render views

LukasGanter (approved):
Bug fixed, works great now!

PR #193, opened 2025-05-18
THO-564 Billboards

IDarkgenesis:
Added Billboards, the function to use a spritesheet will be included in the particle system, since it will be an Addon to the particle system it will be integrated into the billboard as well.

In order to use billboards you need to create a Tag through the menu. And then assign the tag to the component.
image

If one of the components is modified all others with the same tag should be updated. There is the choice to use materials or textures, so just dropping a texture onto the engine is enough to be used and there is no need to create a material for it. Transparencies will come with another PR from Andreu.

To duplicate a billboard either you can duplicate an existing Game Object or create a new one add the component and then select the tag.

The "Delete current tag" will clear all the data of the billboards components and erase the billboard from memory. If a billboard reaches 0 references its gets deleted.

There is a scene in EngineTests/dev-demos with a billboards scene.


caspi99 (approved):
💒

PolCurto (approved):
Works perfect, good job. 
I just found this werid bug, I don't know if it's this specific scene or what, I just upload it to know it exists. I've tested it in other scenes and nothing else disappears so I'll approve it.

https://github.com/user-attachments/assets/a68c926a-6d98-4241-ba8a-6aaaf733b3e8

PR #194, opened 2025-05-18
THO-563 transparent render mode

caspi99:
Alpha Mode: I add a scene with plants from the artists in a branch called ALPHA-MODE in engineTests. You can change to opaque mode or transparent. Also in this branch I solved the problem of crashing when clicking play without a navmesh and you can see a prefab deactivated as another bug solved. (The prefabs when you deactivate them and recharge the scene, it appears activated again)

### DIFF
--- a/Game/EngineDefaults/Assets/Metadata/11_Cylinder0.smeta
+++ b/Game/EngineDefaults/Assets/Metadata/11_Cylinder0.smeta
@@ -23,7 +23,7 @@
             0.0,
             1.0
         ],
-		"defaultMaterialUID": 1300000000000000,
+	"defaultMaterialUID": 1300000000000000,
         "gltfMeshIndex": 0,
         "gltfPrimitiveIndex": 0
     }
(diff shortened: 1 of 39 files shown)


IDarkgenesis (approved):
🪟

LukasGanter (changes requested):
The wireframe rendering is broken with transparent objects

LukasGanter (changes requested):
Wireframe rendering works now, i encountered new problems though: 
- Meshes disappear when moved to certain locations 
SobrassadaEngine_Iy38LXQuwk
- The main character is rendered wrong. The zip file i added contains the 
[walk gltf.zip](https://github.com/user-attachments/files/20327323/walk.gltf.zip)
main character, which is rendered as transparent by default (This only happens with already imported meshes in a project, when newly imported the default setting is opaque. Please make sure that the correct option is selected also for meshes which were imported previously to your changes. After changing to transparent the rendering is broken again. When changing manually to opaque it renders correct.
grafik

SobrassadaEngine_kR2eUAwEGA

LukasGanter (approved):
All bugs fixed, looks great!

LukasGanter (approved):
After the last fix the wireframe rendering now works as well

PR #196, opened 2025-05-19
Task/Level switch-THO-604 

DHazardFenrir:
A script that changes the current scene to another. To test this PR, in EngineTests in the branch test-scene you can load the scene enemiesTesting, selecting one spawn point, delete the script and replace it for the changeSceneScript, add the data (it should be walk in the playerName, and then the name of the scene, for this to work properly to keep the flow of the game, you can do it in navmesh prueba scene, just add navmesh prueba into the scene name) walk to the point and should change scene automatically and go into play mode. 

NOTE: There is NO persistance of data between scene changes, right now if you want to control the player in the new scene, there needs to be a playerprefab and a navmesh set previously. 

This works on release.


SobrassadaEngine/Modules/SceneModule.cpp
```diff
@@ -171,7 +172,7 @@ void SceneModule::CreateScene()
     loadedScene->Init();
 }
 
-void SceneModule::LoadScene(const rapidjson::Value& initialState, const bool forceReload)
+void SOBRASADA_API_ENGINE SceneModule::LoadScene(const rapidjson::Value& initialState, const bool forceReload)
```
PolCurto, line 175: same for these ones

PolCurto (approved):
I tested it in game also, and it works fine, good job! Just take a look at the comments

LukasGanter (approved):
Works, very nice job!

PR #200, opened 2025-05-20
task/FireTrapRandomizer-THO-615

SirKaza:
Things done:
· Trap attack cooldown is random now.
· Gravity can be changed through editor in case we want to fall faster/slower.

I tried using LCG rng from globals but have many problems to implement it to the script because of being a external library so I used srand() standard library. Also, I made in a way that this randomizer between a range of two floats can be used in other scripts.

How to test:
There is still a scene in EngineTests repository and branch dev-testing named FireTrapTesting where this new changes can be tested properly. Try changing min and max attack cooldown times to see that randomizer works and same with the gravity for the speed of the fireball falling.


IDarkgenesis (approved):
:fireworks:

LukasGanter (approved):
Nice

PR #202, opened 2025-05-21
Player improvements

PolCurto:
Improved some of the controls and player features:
- Fixed the player snapping to the axis directions
- Slide along the edges instead of stopping
- When falling after a dash, respawn in the last position after surpassing a y pos threshold
- Move along the camera axes, so moving down the joystick with a rotated camera actually goes down from the player perspective
- Detection of controller connection anytime, not just when opening the engine

### DIFF
--- a/SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/CuChulainn.cpp
+++ b/SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/CuChulainn.cpp
@@ -43,11 +43,21 @@ bool CuChulainn::Init()
     if (!character) GLOG("CharacterController component not found for CuChulainn")
     else speed = character->GetSpeed();
 
-    const GameObject* cameraObj = AppEngine->GetSceneModule()->GetScene()->GetGameObjectByName(cameraName);
-    if (cameraObj && cameraObj->GetComponent<ScriptComponent*>())
+    cameraObject = AppEngine->GetSceneModule()->GetScene()->GetGameObjectByName(cameraName);
+    if (cameraObject && cameraObject->GetComponent<ScriptComponent*>())
     {
-        camera = cameraObj->GetComponent<ScriptComponent*>()->GetScriptByType<CameraMovement>();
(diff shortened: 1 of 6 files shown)


SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/CuChulainn.cpp
```diff
@@ -216,11 +251,18 @@ void CuChulainn::LookAtMouse()
...
+    const float2 stick     = AppEngine->GetInputModule()->GetRightStick();
+    const float3 direction = camFront * stick.y + camRight * stick.x;
     if (direction.LengthSq() > 0.001f) character->LookAt(direction);
 }
 
+void CuChulainn::CheckIsFalling()
+{
+    const float maxDepth = -50.0f;
+
+    if (parent->GetGlobalTransform().TranslatePart().y < maxDepth) SetPosition(lastDashStartPos);
```
SirKaza, line 263: Here, the player should receive 1 damage for falling outside the map / falling parkour jump

SirKaza (changes requested):
Everything works perfectly except falling after a dash, if you test in the tutorial once you fall you respawn in another position and keep falling infinitely.
Not sure if I know what moving along the camera axes is doing, for me, it's the same as it was before. Going down it moves down but not from the camera perspective (it goes diagonally from camera perspective).
Apart from that, excellent job!
Take a look at the comments below.

LukasGanter (approved):
Everything works as expected, the feeling is way better already. Great job!
Take a look at the comments Marc made about the code, apart from that its approved

LukasGanter:
> Not sure if I know what moving along the camera axes is doing, for me, it's the same as it was before. Going down it moves down but not from the camera perspective (it goes diagonally from camera perspective). Apart from that, excellent job! Take a look at the comments below.

If you use the player prefab in the engine test branch, pushing the stick up causes the character to move straight up instead of top the top left. For me this works well.

SirKaza (approved):
Lukas told me that the correct branch was player v2 instead of dev-testing, now it's working for me as well. Good job! Don't forget to fix the comments before merge.

PR #204, opened 2025-05-21
Bug/SomeBugsFixed-THO-625

SirKaza:
Bugs fixed:
- Fixed some warnings in CanvasScalerComponent.
- Stop button was bugging play mode, making it impossible to play.
- Fireball Trap position was wrong, was not using global or local when needed.
- Updated GodModeScript Camera by name.

How to test:
Go to EngineTests repository, branch dev-testing scene FireTrapTesting.
Test pressing stop before playing, and you should be able to play anyway.
Check that trap is still working.
Test adding GodMode script with CuChulainn script and in play mode if pressed P, the player shouldn't be able to do actions and you have free camera and then return with O.


LukasGanter (changes requested):
The firetrap still works, the stop button and warnings are fixed. 
I can't get the script to work though, am i supposed to delete the ChChulain script?

PolCurto (approved):
Works as expected🧯

PR #205, opened 2025-05-22
THO-621 Advanced billboards

IDarkgenesis:
Now transparency is working and also added the option to modify the billboard texture range to just use a part of the whole texture.

billboardRange

### DIFF
--- a/Game/EngineDefaults/Shader/Fragment/BillboardFragment.glsl
+++ b/Game/EngineDefaults/Shader/Fragment/BillboardFragment.glsl
@@ -8,5 +8,7 @@ out vec4 fragColor;
 
 void main()
 {
-    fragColor = texture2D(myTexture, vec2(uv.x, 1 - uv.y));
+    vec4 texColor = texture2D(myTexture, vec2(uv.x, 1 - uv.y));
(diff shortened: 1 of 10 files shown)


LukasGanter (changes requested):
The drag sliders are not exact enough. 
The values need to be able to be set pixel perfect. 
I would make an input for the u coordinate, one for the u width, one for the v coordinate and one for the v height

IDarkgenesis:
> he drag sliders are not exact enough.

You can Ctrl+click the drags and input manually to fine tune the UV coordinates, using x,y and width and height can be tedious since you have to open the texture to figure out those values

LukasGanter (commented):
If the texture is 2048x2048 for example its impossible to enter certain values.
In your gif, if i want coordinate 1000 then i need to add 0.48828125, which is not a valid input. 

Also, regarding your point, you have to know these exact values anyway. You don´t estimate until it looks quite right normally.

PolCurto (approved):
👓 
Only I found a small visual issue, shown in the video (when changing texture the width/height change to the texture size but the texture is repeated)

https://github.com/user-attachments/assets/8c7d357e-09af-4434-9060-1cf4cab09d80

PR #206, opened 2025-05-22
Task/Enemy Archer V1 -THO-571

PabioGF:
First version of the archer enemy.

For test it. Load the EnemyTests Scene on the AcherTest branch of EngineTests.

In the scene you will be able to see the three enemies, including the archer. You can see that it patrols, if he sees the player he follows him and shoots when he is at a medium distance.


LukasGanter (approved):
Works great, nice job!

PR #207, opened 2025-05-22
THO-624 two sided rendering

caspi99:
Testing: Added double side rendering. It has a checkbox in the material for already imported meshes.
Also added a condition in opaque objects to solve the bug of seeing transparent instead of black.
I added the plants as a good example of a double side render. You can enable/disable to check the rendering
[TEST.zip](https://github.com/user-attachments/files/20401154/TEST.zip)


LukasGanter (approved):
Works exactly as expected, nice job!

IDarkgenesis (approved):
Works, nice

PR #208, opened 2025-05-23
BugfixTilefloatUpdate-THO-632

srLlobet:
distance to player now adds world position to the finalPoisiton (localposition). 

You can now input negative values in the start position


LukasGanter (approved):
Works

SirKaza (approved):
Nice changes

PR #209, opened 2025-05-23
THO-633 Scripts disabled after stop playing

caspi99:
Just click play and stop with an enabled script


srLlobet (approved):
ദ്ദി(˵ •̀ ᴗ - ˵ ) ✧

MarcLopezS (approved):
Works perfectly! Well done! 😄 🚀 

Take into account that, when you have a disabled script, if you press play and stop, it automatically enables it. I don't know if this is intented.

MarcLopezS:
Tested with the new commit and the bug mentioned is solved! 👍

PR #210, opened 2025-05-23
Task/spline component THO-43

MarcLopezS:
Now it is available the Spline Component!

With it we can develop different functionalities with scripts such as automated Camera Movement, better patrolling for enemies, etc.

How to test it:
- There is a scene as an example in dev-testing. Open that scene or create a new one from zero.
- In render Options, remember to hit the Spline checkbox in order to always see the splines in the scene. If this is not done, you'll only see them when Game Object containing Spline component is selected.
- (If you have created a new scene from zero) add the spline component and then add some points. Then if you select a point, you can move it with the gizmos. (Take into account that, if a point is selected, **the GameObject Transform and gizmos will be** **deactivated** in order to avoid unexpected behaviour)
- You can modify the position with the modifier section or with gizmos.
- You can delete them.
- If you add a new point, if a point is already selected, the new point will be added next to the selected one. If no point is selected, will be added in the end.

- In properties will take count of the overall points in the spline. Also, alfa changes slightly how open or close the curves are.
- You can hit the checkBox to automatically create a loop in the spline.
- Path probe is to simulate how the path will be when a gameObject will be placed with the spline (The real code with a GameObject will be done in another task). Hit the checkBox to show the marker and you'll see a yellow sphere that goes through the spline and t is the variable that controls the beginning through the end the route of the yellow sphere.

NOTE: Take into account that there is a known bug when two splines are created in the same scene. It is reported in issue [THO-631](https://fpcvideogames.youtrack.cloud/issue/THO-631) Bug when 2 splines in same scene. To sum up, if two different splines have been selected a point, the second spline will replicate the position of all of its point in common with the first spline. At the moment, **don't test it with 2 or more splines**. We'll fix this in the future.

### DIFF
--- a/SobrassadaEngine/Scene/Components/Standalone/SplineComponent.cpp
+++ b/SobrassadaEngine/Scene/Components/Standalone/SplineComponent.cpp
@@ -41,25 +41,28 @@ void SplineComponent::RenderDebug(float deltaTime)
 {
     if (points.size() < 2) return;
 
+    EditorUIModule* ui   = App->GetEditorUIModule();
+    DebugDrawModule* dbg = App->GetDebugDrawModule();
+
     size_t endSeg   = loop ? points.size() : points.size() - 1;
     const float3 curveColor(0, 1, 0); //Green
     const float3 pointColor(1, 0, 0); //Red
 
     auto drawLine = [&](const float3& a, const float3& b)
     {
-        App->GetDebugDrawModule()->DrawLineSegment(LineSegment(a, b), curveColor);
+        dbg->DrawLineSegment(LineSegment(a, b), curveColor);
     };
 
     for (size_t seg = 0; seg < endSeg; ++seg)
     {
         if (loop || points.size() >= 3)
         {
-            float3 prev = EvaluateSegment(seg, 0.0f);
+            float3 prev = EvaluateSegment(seg, 0.0f) + parent->GetGlobalTransform().TranslatePart();
 
             for (int i = 1; i <= stepsDebug; ++i)
             {
                 float u  = (float)i / stepsDebug;
-                float3 p = EvaluateSegment(seg, u);
+                float3 p = EvaluateSegment(seg, u) + parent->GetGlobalTransform().TranslatePart();
                 drawLine(prev, p);
                 prev = p;
             }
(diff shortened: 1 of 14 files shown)


SobrassadaEngine/Scene/Components/Standalone/SplineComponent.cpp
```diff
@@ -0,0 +1,332 @@
...
+    return float3::Lerp(B1, B2, (t - t1) / (t2 - t1));
+}
+
+size_t SplineComponent::Wrap(int i) const
+{
+    const int n = (int)points.size();
+    return (size_t)((i % n + n) % n);
+}
+
+float3 SplineComponent::EvaluateSegment(size_t seg, float segmentT) const
```
caspi99, line 250: const

SobrassadaEngine/Scene/Components/Standalone/SplineComponent.cpp
```diff
@@ -0,0 +1,332 @@
...
+{
+    auto idx = [this](int k)
+    {
+        int n = (int)points.size();
```
caspi99, line 254: const

SobrassadaEngine/Scene/Components/Standalone/SplineComponent.cpp
```diff
@@ -0,0 +1,332 @@
...
+{
+    const float3 worldOffset = parent->GetGlobalTransform().TranslatePart();
+
+    if (points.size() < 2) return worldOffset;
+
+    t = std::clamp(t, 0.f, 1.f);
+
+    const int numSeg = loop ? (int)points.size() : (int)points.size() - 1;
+
+    float segFloat   = t * numSeg;
```
caspi99, line 277: const

SobrassadaEngine/Scene/Components/Standalone/SplineComponent.cpp
```diff
@@ -0,0 +1,332 @@
...
+    if (segFloat >= numSeg) return worldOffset + (loop ? points.front() : points.back());
+
+    int seg = (int)floorf(segFloat);
```
caspi99, line 280: const

SobrassadaEngine/Scene/Components/Standalone/SplineComponent.cpp
```diff
@@ -0,0 +1,332 @@
...
+    float u = segFloat - seg;
```
caspi99, line 281: const

SobrassadaEngine/Scene/Components/Standalone/SplineComponent.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+    if (loop) seg = seg % points.size();
+
+    float3 local = EvaluateSegment((size_t)seg, u);
```
caspi99, line 285: cosnt

caspi99 (approved):
Now its working good, do these little comments and its good to go

PR #211, opened 2025-05-23
Updated fireball trap to work with new hierarchy THO-634

LukasGanter:
Now the fire trap base serves as the main position anchor, so the pivot for moving it around is in the middle of the trap itself. 
Also improved efficency of distance calculation to the player.
For testing, go to dev-test and open the Fire ball scene and check that the pivot for the trap is at the center and the trap still works


SirKaza (approved):
Works perfectly. I tried moving it around and it's robust 👍

PR #214, opened 2025-05-24
Task/Player Character V2 THO-606

PolCurto:
Implemented the player character animations in the game state machine, and made the code adaptations needed for this. Also fixed a bug that attacking in the edge sometimes caused the player to fall. The animations are: idle, run, attack, aim and throw, dash, fall, land, die and respawn. You can tests the repawn one using F6 and F5 to respawn.

KNOWN ISSUE: There is a bug with animations, regarding their playback timer, which cause them to sometimes stop abuptly (for example, you dash and after a few frames the player is in the idle animation while dashing). This is only a visual bug, and while important, it does not break anything (allegedly) and will be fixed in a future task. This way we have animations in this vertical slice.

To test this, I've created a branch in The Hound of Ulster called player-V2, where there is the prefab created. You can place it in a scene and it should work out of the box. 
There is also the scene EnemiesTest in the branch Player-V2 of EngineTests. However, at least in my PC, in this scene the spear projectile is ultra bugged and sometimes works, sometimes not, sometimes appears frozen, anything can happen. You can test it also in this scene, maybe it works for you. In this scene there are also enemies, so you can see the death animation.

Yes, the spear projectile is still a cube.

### DIFF
--- a/SobrassadaEngine/Scene/Components/Standalone/CharacterControllerComponent.h
+++ b/SobrassadaEngine/Scene/Components/Standalone/CharacterControllerComponent.h
@@ -38,6 +38,7 @@ class SOBRASADA_API_ENGINE CharacterControllerComponent : public Component
     const float& GetMaxSpeed() const { return maxSpeed; }
     bool IsDashing() const { return isDashing; }
     float2 GetRealSpeed() const;
+    bool IsGrounded() { return isGrounded; }
 
     void SetTargetDirection(float3 newTargetDirection) { targetDirection = newTargetDirection; }
     void SetMaxSpeed(float newSpeed) { maxSpeed = newSpeed; }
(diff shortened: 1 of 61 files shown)


caspi99 (approved):
🪓

SobrassadaEngine/Scene/Components/Standalone/CharacterControllerComponent.h
```diff
@@ -36,6 +36,9 @@ class SOBRASADA_API_ENGINE CharacterControllerComponent : public Component
     const float3& GetLastPosition() const { return lastPosition; }
     const float& GetSpeed() const { return currentSpeed; }
     const float& GetMaxSpeed() const { return maxSpeed; }
+    bool IsDashing() const { return isDashing; }
+    float2 GetRealSpeed() const;
```
SirKaza, line 40: const &?

SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/Projectile.cpp
```diff
@@ -49,13 +49,12 @@ void Projectile::Shoot(const float3& origin, const float3& direction)
 
 void Projectile::OnCollision(GameObject* otherObject, const float3& collisionNormal)
 {
-    // GLOG("COLLISION IN PROJECTILE with: %s", otherObject->GetName().c_str());
+    GLOG("Collision in projectile with: %s", otherObject->GetName().c_str());
```
SirKaza, line 52: comment this before merge

SirKaza (approved):
Works amazing apart from some bugs that will be fixed as you said in the future.
:shipit:

SobrassadaEngine/Scene/Components/Standalone/CharacterControllerComponent.h
PolCurto: it's a temp value

PR #218, opened 2025-05-24
THO-639 Change Dynamic tree

IDarkgenesis:
Change dynamic tree to be octree.

Load a scene, and everything should keep working


srLlobet (approved):
yes

PR #219, opened 2025-05-24
Task/Enemy Banshee V2 THO-575

PolCurto:
Added the banshee prefab with animations in the state machine and adapted the code for it.

There are almost no code changes. The banshee prefab is available in the branch 'banshee-v2' of the Hound of Ulster repository. It is not placed in any level, to test it load the prefab wherever and get close with the character to see if it works fine. It only has 3 animations: idle, chase and scream. When it flees it uses the chase animation as well.

If something is broken first delete the lib folder just in case


srLlobet (approved):
Also looking very good! GJ

LukasGanter (approved):
Looks very good.
There are times where the behaviour is a bit off, but that might also be because of my player character

PR #220, opened 2025-05-24
THO-565 Decals

caspi99:
I made a branch in EngineTest called Decals. You can put a decal component if you want but needs to face the Z direction. (If you see the decal bad, basically you need to rotate)


PolCurto (approved):
🌳

IDarkgenesis (approved):
Just have a look at the comments sent

PR #221, opened 2025-05-25
Task/basic sound setup THO-612

srLlobet:
Sounds and music update

This adds: 

-all current official ambient sounds for tutorial and level 1.
-placeholder music for tutorial and lvl 1.
-Real time volume, pitch and spacialization control for all sounds (music doesn't have spacialization).
-Distance-based volume attenuation.
-Sounds get disabled when character is too far away so they don't loop while muted.

To test: In engine tests, open the audio test scene. There should be music playing, then, down below a waterfall sound that gets louder when you're closer, and to the left and right there are torch sounds with an octave higher and lower than the waterfall. You can also modify any parameters on the audio sources.


MarcLopezS:
I can't compile. ICE_BLAST is missing.
image
image

PolCurto (changes requested):
Apart from what Marc said, I left you some feedback in discord regarding the audio stop / resume when too far away, which may give problems and right now is not doing anything

MarcLopezS (approved):
Works as expected. Good Job! 😄

SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/ProjectInfo.json
```diff
@@ -0,0 +1,29 @@
...
+   "Name": "SobrassadaSound",
+   "GUID": "{C9341814-2FEF-4729-80F5-E19420EBAC78}",
+   "Generator": "2024.1.3.8749"
+  },
+  "CacheRoot": "../../../../../The-Hound-of-Ulster/SobrassadaSound/.cache",
+  "Platforms": [
+   {
+    "Name": "Windows",
+    "GUID": "{6CA4C6AE-FD2D-42EA-876B-1E8B66830987}",
+    "BasePlatform": "Windows",
```
PolCurto, line 13: This

PolCurto (approved):
Good job. Better remove the ProjectInfo.json from SobrassadaScripts before merging into vs3 🔔

PR #232, opened 2025-05-25
Task/script go spline THO-635

MarcLopezS:
Basic Script to make a GameObject route through a spline.

Condition in order to work:
- the Spline and the GameObject that will contain the script must have the same GO parent. (same level in hierarchy)
- The script cannot be in the same GO as the Spline.

How to test:
- in dev-Testing, the SplineTest scene has been updated for this case. It will contain a GO with a Spline and a GO with the script and a cube mesh. (Remember to activate the Spline CheckBox in Render Options to see the spline at any moment)
- You can play with the speed
- Added PingPong effect: when GO arrives to the end of the spline, it will go backwards and so on.

### DIFF
--- a/Game/EngineDefaults/Assets/Metadata/11_Cylinder0.smeta
+++ b/Game/EngineDefaults/Assets/Metadata/11_Cylinder0.smeta
@@ -23,7 +23,7 @@
             0.0,
             1.0
         ],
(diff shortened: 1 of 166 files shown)


LukasGanter (approved):
Very smooth, nice job!

PR #240, opened 2025-05-30
Task/Mushrooms-THO_652

SirKaza:
Implemented player healing with mushrooms behavior.  
Created a new script named `Mushroom` that contains: a mesh, a script, and a trigger sphere collider.  
When the player collides with the mushroom and presses the healing action, the mushroom disappears, and the player is healed by the amount set in the mushroom.  

**How to test:**  
- Open the `EngineTests` repository, branch `dev-testing`, and load the scene `MushroomTesting`.  
- In this scene, there is a player and two mushrooms.  
- Press the **E** key or **B** button on the gamepad to trigger the healing action.  
- Check the player script to ensure that the player's health is increasing as expected.


IDarkgenesis (approved):
It works, I don't know if its intended but when healingCooldown is set to 0, the healing action does not happen.

SirKaza:
> It works, I don't know if its intended but when healingCooldown is set to 0, the healing action does not happen.

Yeah, it is needed some cooldown, I think is the same with attackCooldown.

PR #241, opened 2025-05-30
Task/DebugVisual-THO_678

SirKaza:
Added debug draw lines for characters (main character and enemies).
Now in render options dropdown there is a new option named Debug Visuals, when pressed it can be seen the front direction of the different characters. This helps also to know here the range attack should be going when aiming.

How to test:
Play in a scene and check that the lines are set and are correct for the different characters.


IDarkgenesis (approved):
Lines are linning.

PR #243, opened 2025-05-31
Task/THO-567 smooth trails

caspi99:
Testing: Go to EngineTests to the branch "Trails". Enter to the scene soldier and activate the attack of the soldier and you can see the trail. If you want to check the trail is in the child of the sword handle inside the right hand bone


IDarkgenesis (approved):
Amazing work! 🚥

PR #244, opened 2025-05-31
Task/Banshee Attack Visual THO_679

SirKaza:
Added some visuals to banshee attack, so the player has more feedback when getting damage.
Visuals are placeholders.

How to test:
Go to HoundOfUlster, branch lg-final-revision and check scene level 2 banshees.


PolCurto (approved):
Works fine. I moved the visuals enabling to when it actually deals damage

IDarkgenesis (approved):
Works good job!

PR #245, opened 2025-06-01
task/MC-combo-implementation-THO-707

PolCurto:
Changes:
* Added a 3 hit combo for the main character. If in the future we want it to be 4 hits, it's only needed to add one more state, the logic remains the same.
* When toggling Debug Visuals checkbox and playing, above the player and enemies there is a debug text with their current state machine state and life.
* Fixed the crash in physics when updating a prefab.
* Little fix in state machine
* Fix an issue with interpolation which sometimes made animations suddenly stop during transitions

Now the attack animation is split in 2: the attack and the recovery. 
The expected behaviour is: while inputting attack, only play the attack animations. When the tird attack is reached or the player no longer attacks, the recovery animation plays. You can't move while in a combo, but you can dash after an attack animation and continue the combo afterwards to reposition yourself, avoiding the recovery animation.

To test: In TheHoundOfUlster repo there is a branch called task/player-combo. Play there and check if it behaves as expected.


SirKaza (approved):
Agree with Tomeu, recovery should be superfast to not force players to dash after finishing a combo.

PR #246, opened 2025-06-03
Task/Health-hud-connection-THO-648

jmartiolivart:
The life connection to the HUD is working.
- When die -> Full life in HUD
- When soldier attack or when mushroom is eating -> Update health HUD

To try change:
     float healCooldown                          = 0.1f; 
in _Character.h_ to make the mushroom work.

Go to SCENE_Tutorial of merge/lg-merge-revision and you can try:
- using the first mushroom (clicking E).
- fighting the soldier and diying.


PolCurto (approved):
Works fine, good job. Just merge and good to go

PR #248, opened 2025-06-04
Task/navmesh config save THO-661

srLlobet:
To test: Load any scene with a navmesh, and open the navmesh metafile, check that the values in the editor are the same.

only rcConfig values, the values that are not stored in the metafile aren't kept. The UI default values have been fixed to always use the best ones, which should be kept across all navmeshes. These are the WATERSHED instead of MONOTONE partition type, and the agent parameters which dont do anything right now.


IDarkgenesis (approved):
👍

PR #249, opened 2025-06-05
task/MC special attack THO-680 & THO-718 player inputs update

PolCurto:
Implemented the base of the character ultimate attack. Also changed the inputs to the new proposal

Press F (keyboard) or B (controller) to perform the ultimate attack. Right now it is an area that deals 2 damage to enemies inside. The range, damage and cooldown has to be iterated, the current values were just for testing. Also the animation is a placeholder (basically the fall animation).

There is a branch called task/ultimate-attack in TheHoundOfUlster repo with the new state machine and player prefab to test it.

It is in consideration that the ultimate attacks maybe can also root enemies, I think this is better to do in a separate task, and develop there a way to root enemies (and player maybe) in general. Also now it has a normal cooldown, in the future maybe it also gets recharged with breakable objects.


SirKaza (approved):
Well done, all works as desired!
I think it will be better to use ultimate with E in keyboard instead of F and use F to interact/activate things, but we can discuss it.

PR #252, opened 2025-06-06
Task/Animation speed parameter-THO-656 

DHazardFenrir:
This PR allows the speed of the animation be controlled independently in the state machine editor. 

How to Test: Open the Hound of Ulster repo, go to tutorial (where the prefab of player is), select player, open de state machine editor and load the player state machine, in the different states should be a slider with the name animation speed, tinker with it and put the states with different speeds. You can check how the speed is affecting the animation by  pressing the triggers, the transition should look good and the animation speed remain independent in both clips.


PolCurto (changes requested):
Works fine, however this value is not stored. If I modify the speed of some clips, close and open the engine again, they are reset to 1

DHazardFenrir:
> Works fine, however this value is not stored. If I modify the speed of some clips, close and open the engine again, they are reset to 1

I edited the statemachine manager so you can save the values now, but you need to re-save the state machine. Right  now, since the state machine existed before  this upgrade, the clip animation will load at 0 speed since the value didn't exist before. If you make a new state machine, then It will start at the default speed of 1.

PolCurto (approved):
Perfect, good job!

PR #253, opened 2025-06-07
Task/RedoPauseMenuScript-THO-711

jmartiolivart:
Redone the PauseMenuScript and added pause and continue of the game when in Pause Canvas.
I also added 0 life when dying with the correct texture in this same branch.

How to check?

* Go to Scene_Tutorial of Hound of Ulster repo and try the pause menu if it's working correctly


SirKaza (dismissed):
Good job!
Two things that needs to be improved:
- When hit continue from menu pause, player dashes always. 
- When hit go to main menu, it should go directly to new adventure - options - exit and not to press any key.

MarcLopezS:
I've checked and what Marc commented, this is solved:
- When hit continue from menu pause, player dashes always.

And this is not:
- When hit go to main menu, it should go directly to new adventure - options - exit and not to press any key.

PR #254, opened 2025-06-08
task/New banshee behaviour THO-720

PolCurto:
Implemented the new banshee behaviour as it was discussed. Now the banshee no longer flees. When the player is detected, it chases it until a set distance, then disappears and after a random amount of time appears next to the player, attacks, disappears again, attacks, etc. If the player gets far it is chased again or even stops chasing it.

The changes on shaders and materials do nothing for now, I tried some things with Andreu to make the banshee disappear smoothly tweaking the alpha, but there were more problems than we expected, so for now it directly disables and enables the mesh. That will be done in a future task.

To test you can go to the level 2 of the game repo and see if it works fine. Raise chase range and the attack range of the banshee, so it disappears further away from the player.


DHazardFenrir (approved):
Looks good! This is more about gameplay because I think the movement is a bit slow, and I can dodge them quite easily even with the teleport, but the behavior is working as expected.

PR #256, opened 2025-06-09
Task/spline game object rotation tho 663

MarcLopezS:
The Spline now incorporates rotation of the splinePoints and speed variable that indicates at which velocity the GameObject shound arrive at that point. 

The spline points have been changed for cones, in order to show the rotation in which the gameObject should have.

In order to test it:
- There is a scene created in dev-testing called "SplineTestV2". You can use that or create a spline from zero. (Take into account that if you decide to create it from zero, the GO that contain the spline has to be a sibling of the GO that would contain the script to move through the spline).
- Go to the spline and tweak the position and rotation values (scale is ignored), You can use gizmos to modify it easily.
- Set a speed value in each point. This speed variable is interpolated. In order to change the speed of the GameObject, that is modified in the script MoveGOInSpline inspector.
- Go to play mode and tweak also the parameters mention above to see how it reacts.
- Try it with a loop and non-loop spline.
- Tweak the pingpong parameter (only for non-loop splines). Is located in the script component for MoveGOInSpline script.

(NOTE: there are three functions that almost have the same code but with different parameters according to the values requested (Evaluate, EvaluateRotation & EvaluateSpeed). I'm thinking of a way to adapt them to not repeat code. I'm making the pull request now to test the functionality that works correctly and then modify that part to make it more "clean").

### DIFF
--- a/SobrassadaEngine/Scene/Components/Standalone/SplineComponent.cpp
+++ b/SobrassadaEngine/Scene/Components/Standalone/SplineComponent.cpp
@@ -173,7 +173,7 @@ void SplineComponent::RenderEditorInspector()
     ImGui::SeparatorText("Path Probe");
 
     ImGui::Checkbox("Show marker", &showMarker);
-    if (showMarker) ImGui::DragFloat("t  (0-1)", &markerT, 0.001f, 0.f, 1.f, "%.3f");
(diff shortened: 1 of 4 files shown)


PolCurto (approved):
Works fine, good job! One thing I've noticed is that if either the spline or the cube with the script are children of another object, it doesn't seem to work. However I don't know if this will be a problem

PR #257, opened 2025-06-10
task/Prefabs usability upgrade THO-638

PolCurto:
The main change of this PR is how prefabs are overriden. Now prefabs are no longer deleted and created again, but simply updated, so the gameObjects and components UIDs remain the same. This allows some features like the gameObject field in scripting be useful again. Also, because they are not deleted they can be children of other gameObjects. I'm pretty sure this change also fixes some other problems there could be with batching and other things, but I have not checked.

However, this only works with new prefabs, because a new field has been added to the prefab files. Old prefabs are still overriden the old way, but once they are overriden, they should be updated and start overriding in the new way. Eventually all prefabs that already exist in the game will probably get updated wihtout needing to recreate them.

I also added a new field prefabVersionUID, which keeps track of the prefab version. This way, prefabs are only overriden if they have changed, so when loading a new scene no prefabs will be overrriden except if they have been updated in another scene. This also works only with new prefabs.

TO TEST: go in the game repo, modify prefabs and load scenes to check if everything works fine. There are some logs that say how many prefabs have been overriden, if in the old or new way, etc., to help see if it's correct.

This task had another part of saving individually modified prefabs when saving the scene, but I have not done it here. I don't even know if that's needed. If you want to save a modified prefab, you can unlink it before saving the scene and it will be properly saved.

### DIFF
--- a/SobrassadaEngine/FileSystem/PrefabManager.cpp
+++ b/SobrassadaEngine/FileSystem/PrefabManager.cpp
@@ -16,7 +16,7 @@
 
 namespace PrefabManager
 {
-    UID SavePrefab(const GameObject* gameObject, bool override)
+    UID SavePrefab(const GameObject* gameObject, bool override, const UID versionUID)
     {
         // Create doc JSON
         rapidjson::Document doc;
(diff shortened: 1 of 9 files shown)


IDarkgenesis (approved):
Looks good, nothing seems broken. 👍

caspi99 (approved):
✌️ Looks good

PR #258, opened 2025-06-11
task/Archer arrow projectile name THO-714

PolCurto:
Just a small change which makes each archer get their own arrow reference, without needing a name. To test, play a level which has archers and test if they shoot properly as before.


PabioGF (approved):
Works perfect, Great! 🥇

PR #259, opened 2025-06-12
task/God mode player THO-664

PolCurto:
Now, when you are in the god camera while playing, you can press the key 'I' to teleport the player to the camera position (The god camera was enabled with 'P' and disabled with 'O'). You can also change the fov for the god camera.

Also, now the player script has a GodMode checkbox, which if enabled the player can not take damage, to avoid dying while something else is being tested. This can also be enabled and disabled with the F7 key, so it can be used in the game build as well.


MarcLopezS (approved):
Apart from the const doubt, it works smoothly! Great job! 🥇

PR #262, opened 2025-06-14
task/Enemy behaviour update THO-737

PolCurto:
Now enemies have a new field called maxDetectionRange. If the player is further than this, the enemy will stop chasing. It will stand still searching for the player (right now is the idle animation, would be nice to be an animaion where it looks around), and after the duration delcared in the new field searchDuration the enemy goes back to patrol (the banshee just goes to the start position, it will teleport when merged with the banshee PR).
Also, if the player is dead enemies go back to patroling.

TO TEST: Just play in some levels and test if the differend types of enemies behave as expected, stopping to chase if the player is too far or dead.


PabioGF (approved):
Works Perfect, great Job!

PR #263, opened 2025-06-14
THO-562 Particle System

IDarkgenesis:
Firist you need to create a ParticleSystem, and then once assigned you can create different emitters and customize the different parameters by using addons.

There is a branch in engine test with a fire as an example (particleSystem), also I have been testing and working with some artists to check how it was working.

PD: ALSO ADDED CTRL+S  as hotkey to save the scene

labubu


SobrassadaEngine/Modules/ParticleSystemModule.cpp
```diff
@@ -0,0 +1,134 @@
...
+{
+}
+
+ParticleSystemModule::~ParticleSystemModule()
+{
+}
+
+bool ParticleSystemModule::Init()
+{
+    float vertexData[] = {
```
caspi99, line 22: Context expression

Game/EngineDefaults/Shader/Vertex/SpritesheetVertex.glsl
```diff
@@ -0,0 +1,43 @@
...
+layout(location=5) uniform float currentFrame;
+
+out vec2 uv;
+out vec2 uvNext;
+flat out float blendFactor;
+
+void main()
+{
+
+    float X = trunc(mod(currentFrame + tileOffset.x, tileSize.x));
```
caspi99, line 22: Check const in shader

SobrassadaEngine/Modules/ParticleSystemModule.cpp
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    glDeleteBuffers(1, &quadVBO);
+
+    return true;
+}
+
+void ParticleSystemModule::RenderParticles()
+{
+    bool playMode                     = App->GetSceneModule()->GetInPlayMode();
```
caspi99, line 62: const

caspi99 (approved):
Goog job with the particles, just check the comments I left here and its good to go

PR #266, opened 2025-06-17
fixed navmesh valid propagated to children THO-630

MarcLopezS:
Now the navMesh valid is propagated to the children.

How to test it:
- Test it with any gameObject that has childs.
- When you check the NavMesh Valid, it should work like the enable check. If you activate the checkbox, all children should activate. Also happens when deactivate.


IDarkgenesis (approved):
Works as expected, good job and thanks for the changes!

PolCurto (approved):
👍

PR #272, opened 2025-06-19
THO-747 trails seeing camera

caspi99:
Now the trail is always seeing the camera. 
Testing: Put a trail and a spline and check if always is seeing the camera


IDarkgenesis (approved):
📷

PR #274, opened 2025-06-20
task/Player charged attack THO-742

PolCurto:
New features:

- Charged attack for the player. Keep the attack button pressed to charge, and release to do a 360 area attack. If not charged enough, the character will simply stop charging. If an enemy attacks you, the charge is interrupted. There is no visual feedback of this because there is no "hurt" animation, but the charging timer is reset.
- Implemented the new player animations: running, combo attacks, special attack and charged attack.
- Fixed the ranged attack spam bug.
- New debug visuals for the player to show also the logic state. Made easy to add more logs to any enemy that needs it.
- Now with each attack the player moves slightly forward, moving a little further with the last attack of the chain. You can also change the direction of the attacks mid-combo with the controller.
- Slightly tilting the joystick makes the player move slower (The animation is the same).

TO TEST: Go to the alpha1 branch of the game repo, where the new player prefab and state machine are, and play (remember to delete the library). Check if the added features work as expected, and spam buttons like a madman to see if the character eventually breaks. 

The aim animation may do weird things. That's because right now the animation is being paused and resumed when shooting, in the future it will be 2 separated animations to ensure they are properly displayed


caspi99 (approved):
Works fine! The only thing is that when an enemy attacks you (a consideration) to have a timer or something to know that is reset

SirKaza (approved):
Nicely done, before accepting this pr merge to solve conflitcts and check alpha1 in houndOfUlster cause charge attack is not working there.

PR #277, opened 2025-06-22
Bug/dash bug fixes THO-750

PolCurto:
Note: This PR is built over the Charged attack one, if that's still open test that one first.

Fix dash snapping back when dashing a little over the edge of a cliff.

To test: test that dashing over the different drops of the game works fine (for example the one in the tutorial main area or the level 1 first checkpoint with no return). The falling speed has also been increased.

### DIFF
--- a/Game/EngineDefaults/Shader/Fragment/ParticleSystemFragment.glsl
+++ b/Game/EngineDefaults/Shader/Fragment/ParticleSystemFragment.glsl
@@ -6,6 +6,7 @@ in vec2 uv;
 in vec2 uvNext;
 flat in float blendFactor;
 flat in vec4 fragParticleColor;
+flat in float fragColorIntensity;
 
 out vec4 fragColor;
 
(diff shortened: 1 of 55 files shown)


LukasGanter (changes requested):
Works definitely better than before, in my mind at the ledge in the tutorial the player should slide down sideways like it does at a wall instead of stopping at the ledge.

PolCurto:
Now the dash constantly updates if its destination is in a navmesh and the coliisions check, so if it's only colliding with a lateral raycast the character moves a little to the other side to avoid situations like that one, and overall feels better. With DebugVisual enabled you can see the three raycasts of the dash, yellow if no obstacles and red if obstacles.
This change, however, has a bit of impact in performance because of the raycasts each frame (I've noticed about 1-2ms increase in level 2, no increase in the tutorial, and 1 ms in level 1, I guess because of the number of elements in each level). 
There is a checkbox in the CharacterController component to enable/disable this new "precise" dash, in case we need the old one, not only for performance but because maybe this new one has some bugs I've not found.

PR #279, opened 2025-06-22
Task/camera shake THO-748

PolCurto:
Added the camera shake functionality to the camera script. You can call the function StartShake with a duration, an intensity and optionally a smoothness, and the camera will shake accordingly.

To test: play and get hit by enemies, and see if the shake works. You can change its values in the CuChuainn script to see the differences.


SirKaza (approved):
Well done!

IDarkgenesis (approved):
🫨

PR #282, opened 2025-06-23
Task/Enemy-Spawner-THO-619

jmartiolivart:
How to test:

I created a Scene in the hound of Ulster in the branch named task/EnemySpawner.
You can test it with the SpawnerEnemy GO that's alerady created or create a new one and adding the EnemySpawnerScript and CubeCollider with Trigger and layer trigger.
Passing the UID of the enemy prefab it generates new enemies inside the collider.
The number of enemies can be modifiend in the inspector as well as the enemy.

Prefabs UID i tested with:
1676077673814985 -> Archer.
1652372403612629 -> Soldier
1646438127870003 -> Banshee

There's a little lag when are being generated.
Please tell me if it needs to be changed something.

https://github.com/user-attachments/assets/37c67092-990d-4155-8e60-46c8fed0f3c1

### DIFF
--- a/SobrassadaEngine/FileSystem/PrefabManager.h
+++ b/SobrassadaEngine/FileSystem/PrefabManager.h
@@ -13,5 +13,5 @@ namespace PrefabManager
     void CopyPrefab(
         const std::string& filePath, const std::string& targetFilePath, const std::string& name, const UID sourceUID
     );
-    ResourcePrefab* LoadPrefab(UID prefabUID);
+    SOBRASADA_API_ENGINE ResourcePrefab* LoadPrefab(UID prefabUID);
 } // namespace PrefabManager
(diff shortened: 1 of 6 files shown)


LukasGanter (changes requested):
Basic functionality works as expected, some changes are necessary though:
- When the spawn Once option is not checked, the script spawns enemies continuously, leading to a crash after a couple of seconds. Maybe implement it so that once the player left and reentered the trigger box the enemy gets spawned again
- Can you change it so the trigger box is a component of a child game object while keeping the spawn location as the parents GO? That way, we can move the trigger to somewhere else and trigger the spawn from another place
- The lag when spawning the enemy is a bit of a problem. I´m guessing its because of the dynamic octree update. Can you try to reduce that lag a bit?

LukasGanter (approved):
Works as expected now, good job!

PR #283, opened 2025-06-23
Task/Soldier knock back V1 -THO-754

PabioGF:
A first version of the knockback when a soldier is hitted.

For testing open any scene with soldiers in any branch of the game repo 

Now when you hit an enemy they stop what they are doing for having a knockback in the direction it was hitted. 
The animation changes to the idle one but it will be changed to a hitted one in the future.
The distance of the knockback and the time the enemy is stunned can be changed in the inspector.

I'ts a first version that can be improved when we have more time after alpha 1. Any improvement ideas are welcomed.


PolCurto (approved):
Works fine, good job. I changed a little the knockback values so it feels better with the player combo speed and movement

PabioGF:
> Works fine, good job. I changed a little the knockback values so it feels better with the player combo speed and movement

Perfect, Thanks!

PR #285, opened 2025-06-23
Bug/THO-761 decals disabling bug

caspi99:
Now you can enable/disable a decal.
Also added an option to not use curves in trails and solved the problem with alpha.


SirKaza (approved):
Good work!

PR #287, opened 2025-06-24
Task/Add-Gameover-screen-THO-745

jmartiolivart:
How to test:

Go to task/EnemySpawner branch in the Hound of Ulster.
When dying it should appear gameover screen (it's the same Image as was the PauseMenu but it's actually the GameOver Image).
It should stop the game.
It should let you go to MainMenu or to Respawn with full health.

https://github.com/user-attachments/assets/b6197711-64e9-4c4e-8cbb-9020a07efa27


DHazardFenrir (changes requested):
It works with the continue option, but when I tried to press the main menu option it didn't show anything (im playing with keyboard). And just for a note, the font could be a brighter, not black, since i had problems to see it with my laptop brightness and I had to increase it a little bit to see the letters clearly. 
gameOver

PR #292, opened 2025-06-25
Camera movement bug

LukasGanter:
Fixed the camera movement when the framerate drops too much / delta time increases too much. 
Previously happened when loading a new scene during play or when playing on a low performance pc.

For testing, you can use the following code to artifically increase frame duration (Just put it somewhere in the update loop): 
std::this_thread::sleep_for(std::chrono::milliseconds(15));
You should see a varity of bugs including the player falling through the floor and it not being able to move properly. Those issues will be adressed later. What should work is the camera, it should stay focused on the player all the time.


SirKaza (approved):
Tried with different low frame rates and camera bug is fixed.

PR #293, opened 2025-06-25
Tho 757 go update culling

IDarkgenesis:
FIRST, DO NOT MERGE THIS PR BEFORE ALPHA_1 DELIVERY.

This PR contains several features, mainly only objects inside a given area and the camera frustum are updated and rendere, so we don't waste time calculating logic and animations for the whole scene. It also includes:

- OnCollisionEnter / Exit for colliders.
- Tag system (and add up to 10 tags to a game object and get all game objects that containg a given tag).
- Player location tag and script to manage the player location.

With those changes we will need to create a trigger collider and asign the PlayerLocationScript to it, once a tag is set in the script and the desired gameObjects they will update even if they are not inside the frustum.

HOW TO TEST:
1. Go to task/updateCulling branch in Hound of Ulster.
2. Delete Library.
3. Play through the level.

Comparison of performance in Level2 (worst case scenario):

image

With those changes:

LEVEL_2_UPDATE_CULLING

### DIFF
--- a/SobrassadaEngine/Modules/PhysicsModule.cpp
+++ b/SobrassadaEngine/Modules/PhysicsModule.cpp
@@ -141,10 +141,6 @@ update_status PhysicsModule::PreUpdate(float time)
                     firstUserPointer.onCollisionExitCallback->Call(
                         secondUserPointer.collider->GetParent(), secondUserPointer.layer
                     );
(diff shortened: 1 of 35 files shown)


LukasGanter (changes requested):
In the game release with this engine version there are a couple of problems, mainly with the ui and enemies. The soldier in the tutorial for example doesn´t move for me

PolCurto (approved):
🎲

PR #295, opened 2025-06-25
UIDs changed for cooldown and options menu for alpha1

MarcLopezS:
Updated the references of the UI for the cooldown and the UI controller Menu for alpha1 in The Hound of Ulster Repository.

In order to test this:
- make a dash or ultimate attack on level 2 and see if it works the change of the icons of the skills.
- in options menu (main menu scene), check that changes the button navigation of the menu when using a controller and changes back when using keyboard.


SirKaza (approved):
It works smoothly!
In options in main menu, why game pad images are better than keyboard though?
They have a white touch in the background that keyboard doesn't have.

LukasGanter (approved):
Looks good and works well

PR #302, opened 2025-06-26
Feature/ambient occlusion THO-659

srLlobet:
Adds ambient occlusion to the engine!

The main updates are 2 new render passes with 3 new shaders, the first one is used for ambient occlusion, the second one is just the debug, and the last one is a blur filter applied over the first one, to make it smooth.

The way it works is the SSAO class generates a set of random tangents (set to 32 for now, which is medium-quality) taken from the top hemisphere of a sphere drawn on every pixel, with a bias on the center of the sphere, then the shader grabs those tangents and, with a randomized noise texture, and the gbuffer, tries to find occluded geometry, and applies a grayscale on the resulting image. Making the borders of objects way more clear.

After that, a blur filter is applied twice, first horizontally, then vertically, to make sure there's no patters on the darker areas, and increase the smoothness. Then it's passed to the lighting pass.

To test: Open any scene, look at the SSAO, it's beautiful. 

There is a debug mode in which you can see the SSAO texture with the blur applied. But it doesn't update when you move, because I'm too tired to figure it out, and it doesn't change anything, just click on and off when you move the camera.

imagen


caspi99 (approved):
The ambient oclussion is working, the only thing is with the SSAO checkbox, is not updating when the camera is moving, you should check that. Good work anyway

caspi99 (commented):
Also check the comments I left for you

PR #304, opened 2025-06-28
Fixed the MainMenuSelectorScript logic and added layer in canvas

jmartiolivart:
Fixed the logic in MainMenuSelectorScript. It now works properly with both GameOverScript and PauseMenuScript.

Added a variable to Transform2D to control the draw order of elements in the Canvas (default is 0). The higher the number, the closer the element appears (it is saved and loaded properly). This prevents UI overlapping issues.

To test:
Go to The Hound Of Ulster repository, branch alpha_1, and test it in the SCENE_TUTORIAL. If it works correctly, I’ll add the prefabs to the other scenes. (I’ll recreate the changes in hound of ulster repo once the other PR removes them.)

### DIFF
--- a/SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/PauseMenuScript.cpp
+++ b/SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/PauseMenuScript.cpp
@@ -12,47 +12,70 @@
 
 bool PauseMenuScript::Init()
 {
+    CachePanel();
+    if (cachedTarget) cachedTarget->SetEnabledRecursive(false);
+    isOpen = false;
     return true;
 }
 
-void PauseMenuScript::Update(float deltaTime)
+void PauseMenuScript::Show()
 {
-    const KeyState* keys           = AppEngine->GetInputModule()->GetKeyboard();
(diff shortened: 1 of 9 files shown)


PolCurto (approved):
Works, good job. Sometimes UI images still cover others behind them even if that part is transparent, so changing their z value will still be needed

SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/PauseMenuScript.cpp
```diff
@@ -12,47 +12,70 @@
...
+    const KeyState* k = AppEngine->GetInputModule()->GetKeyboard();
+    if (k[SDL_SCANCODE_ESCAPE] == KEY_DOWN) Toggle();
+}
+
+void PauseMenuScript::Save(rapidjson::Value& out, rapidjson::Document::AllocatorType& a)
+{
+    out.AddMember("PanelToShow", rapidjson::Value(panelToShowName.c_str(), a), a);
 }
 
-void PauseMenuScript::Save(rapidjson::Value& targetState, rapidjson::Document::AllocatorType& allocator)
```
caspi99, line 47: Update the script behaviour, save is not longer existing and you dont need to include a load function, its all done automatically

PabioGF (approved):
Seem's fine, good job!

PR #306, opened 2025-07-04
THO-844 crash on changing scene

caspi99:
Solved the crash when changing scene and when clicking play in debug mode. Also added search bar in gameobject field.


srLlobet (approved):
All good!!!!

PR #308, opened 2025-07-07
task/Resource UID as script parameters THO-770

PolCurto:
Added a new script field of type Resource, which stores the UID of a resource to be used in the script. In the inspector of a script, it is a dropbox whith all the existing resources the library has. You can search the one you want introducing the name, like in the gameObjects field. 

To test: open the engine and in the CuChulainn script there are 4 new fields with the images of the HUD ultimate and dash, you can fill those fields with the resource textures you want and see if they show properly

### DIFF
--- a/SobrassadaEngine/Modules/ResourcesModule.h
+++ b/SobrassadaEngine/Modules/ResourcesModule.h
@@ -8,7 +8,7 @@ class BatchManager;
 class GeometryBatch;
 class Resource;
 
-class ResourcesModule : public Module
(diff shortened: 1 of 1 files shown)


srLlobet (approved):
Allright! Code looks pretty good and it works nicely! GG

IDarkgenesis (approved):
Works, good job!

Since now we would be able to request resources from scirpts the SOBRASADA_API_ENGINE can be added to the ResourcesModule so we can acces it from scripts.

PR #309, opened 2025-07-08
Task/THO-728 light culling

caspi99:
Testing: You can play the game and check if point lights and spot lights are working correctly. Also you can add, delete, move and change parameters to chech all is properly working. Also added icon for the engine.


IDarkgenesis (approved):
Ligths still work, good job.
🍨

PR #310, opened 2025-07-11
bug/RenderPopping-THO-653

SirKaza:
Fixed popping with game objects that were enabled disabled frequently.
There is still an issue with the trail of the range attack.

How to test:
Try using range attack, ultimate and see that for a frame there is not an incorrect position of these meshes.

### DIFF
--- a/SobrassadaEngine/FileSystem/Batching/GeometryBatch.cpp
+++ b/SobrassadaEngine/FileSystem/Batching/GeometryBatch.cpp
@@ -252,8 +252,16 @@ void GeometryBatch::GenerateCommands(const std::vector<MeshComponent*>& meshes, 
     totalVertexCount = 0;
     totalIndexCount  = 0;
 
-    for (const MeshComponent* component : meshes)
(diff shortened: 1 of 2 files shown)


IDarkgenesis (approved):
Havn't seen any popping with this changes, good job!

🧯

SobrassadaEngine/FileSystem/Batching/GeometryBatch.cpp
```diff
@@ -252,8 +252,16 @@ void GeometryBatch::GenerateCommands(const std::vector<MeshComponent*>& meshes,
     totalVertexCount = 0;
     totalIndexCount  = 0;
 
-    for (const MeshComponent* component : meshes)
+    for (MeshComponent* component : meshes)
     {
+        if (!component->GetWasEnabled() && !component->GetBatchWasEnabled())
+        {
+            GLOG("%s", component->GetParent()->GetName());
```
IDarkgenesis, line 259: Maybe delete this GLOGS before merge

PR #311, opened 2025-07-13
task/Riastrad basic behaviour THO-777

PolCurto:
Implemented the riastrad mode basic behaviour. There are no new VFX or animations for now, that will be added in a future task when artists have it done. It is triggered with Q or LT + RT.

Now, when triggering the riastrad, animations are speed up, dash cooldown is reduced, charged attack charges twice as fast, you move faster, are invulnerable and heal to max health. I don't know if the idea is also to deal more damage, it can be easily changed. You can transform in the middle of a combo, but not in the middle of an action. Also, when riastrad finished it waits for the player to finish the current action, or even to finish the combo before returning to normal state.

TO TEST: Play any level and see if it charges and behaves as expected. With Debug Visuals you can see the riastrad meter and whether it is enabled or not. You can also press F8 to fill it immediatly to test. There are some riastrad parameters in the CuChulainn script you can tweak.

I also did some changes to fill the bar in the HUD, however that will probably be replaced with a shader when the custom shader pipeline is done, and I don't want to create conflicts in the scenes, so it is not placed in any scene to test.


IDarkgenesis (approved):
Tested and works as expected

PR #312, opened 2025-07-14
bug/Materials properties saving THO-854

PolCurto:
Fixes:
- Now materials properly save & load their properties.
- Fixed that changing the diffuse color of a material only worked in transparent meshes, now works also in opaque and alpha.
- Changing a property of a material updates all the meshes in the scene that use the same resource material 

TO TEST: tweak materials parameters, save and load scenes, and try deleting the library folder to see if it works properly.


caspi99 (approved):
⭕

PolCurto:
> Looks like all parameters are updated for the material but the render mode, I don't know if its intended, even then good job!

It's because renderMode belongs to the meshComponent, and there is already a lot of rendring logic that depends on that, so I decided to leave it like this

PR #313, opened 2025-07-20
Task/THO-787 emissive material

caspi99:
Added emissive material to materials.  Testing: You can add meshes with emissive material and see if it working. I leave here a mesh from Victor that has emissive to check.
[monolith.zip](https://github.com/user-attachments/files/21335315/monolith.zip)


IDarkgenesis (approved):
Works, good job!

🌾

PolCurto (approved):
Nice 🪹

PR #314, opened 2025-07-23
THO-753 Custom shader render pipeline

IDarkgenesis:
This PR contains the ShaderScriptComponent, which has 3 different settings in order to set where the render script should be called in the render pass. It also fixes an issue with the PS that made it so there were rendering issues and some would stop rendering, load Tutorial and go to altar, all candles should be lit now.

There are 3 MovingUV shader scripts as example, as well as a normal script that switches between two render scripts, hit key 6 and 7 to switch between them. There can be more than one shader script enabled, just make sure it makes sense that more than one are active at a time.

PLEASE DO NOT USE THE RENDER FUNCTION INSIDE NORMAL SCRIPTS.

To test: Load EngineTests repo and the shaderScripts branch.

movingUV_full


caspi99 (approved):
♿

MarcLopezS (approved):
🔥 🔥 🔥 🔥 🔥 🔥

PR #315, opened 2025-07-24
Breakable objects THO-833

LukasGanter:
Added a script for breakable objects.

For testing, checkout the same named branch in the game repository and add the prefabs named "Destructible_..." to the scene. 
When playing, these objects can be destroyed with melee or range attacks


PolCurto (approved):
It wasn't compiling at frist, but I made a change in the .vcxproj file that seems to fix it. Apart from that, works as expected, good job!

jmartiolivart (approved):
Destructible objects break as expected with melee, ranged, and ultimate attacks. Cool effects.
The prefabs show up without materials on my side, and their collider doesn’t interact with the player, fine if that’s intentional; otherwise we can tweak it in the editor.
Everything else looks good. Nice work!

PR #316, opened 2025-07-26
task/Player first SFX and Fix Collisions bug

PolCurto:
This PR adds the following sound effects to the player: attack, receive damage, dash & steps. For the steps, a raycast is thrown to the ground to get the tag of the gameObject, to play a sound depending on the material below.
Also I fixed a weird bug where the player attack didn't generate collisions, at least on release. 

TO TEST: There is a branch called player-SFX in theHoundOfUlster repo. You need to add this attached soundbank for the SFX to play in Assets/Soundbanks/Windows. 

The tags for the steps sound are only placed in the tutorial, when this gets accepted it will be placed everywhere. Also, when adding this new soundbank the music won't play, that's also fine because the default events are reset, it will be added again also on approval.

[Soundbank](https://drive.google.com/file/d/1LXX-4wmFbeNqHuFqckPJy6TtmZwHHcLB/view?usp=sharing)


jmartiolivart (approved):
Tested the branch and everything works as expected. All SFX (attack, damage, dash, and steps) trigger correctly. The step sounds based on ground material works perfect.

When I move slowly walking the steps doesn't sound i don't know if it is intended but aside from that, everything feels solid. Nice work!

IDarkgenesis (approved):
Tested the changes and didn't find anything weird or broken, good job!
 👍

PR #320, opened 2025-07-28
Task/Soldier new animations

PabioGF:
First iteration of the soldier final version task.

New animations of the soldier implemented. For seeing them go to the Soldier-final-version branch of the Game repo and play.

You can see how now doesn't automatically disappear when it dies but it plays its animations before doing it and also how it plays an animation when he detects the player. Also has a different idle animation for doing it between attacks.

Don't delete this branch when accepted

### DIFF
--- a/SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/Soldier.cpp
+++ b/SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/Soldier.cpp
@@ -50,7 +50,13 @@ bool Soldier::Init()
 
 void Soldier::Update(float deltaTime)
 {
-    if (agentAI == nullptr) return;
+    if (currentState == SoldierStates::DEATH)
+    {
+        deathTimer += deltaTime;
+        if (deathTimer > 3.0f) parent->SetEnabled(false);
+    }
+
+    if (currentState == SoldierStates::DEATH || agentAI == nullptr) return;
 
     if (isKnockback)
     {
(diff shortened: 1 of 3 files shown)


PolCurto (changes requested):
There are some issues with the soldiers, at least in level 1, I sent you a video through discord

PolCurto (approved):
Fixed

SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/Soldier.cpp
```diff
@@ -114,6 +125,9 @@ void Soldier::OnDamageTaken(int amount)
     isKnockback    = true;
     knockbackTimer = knockbackTime;
     ApplyKnockback();
+    HashString animStateFromPlayer = GetAnimStateNameFromPlayer();
```
LukasGanter, line 128: This function is not found, therefore the build fails for me

LukasGanter (changes requested):
Build fails, see comment

PabioGF:
Fixed, that line is from a future work, it got through the push

LukasGanter (changes requested):
Dying animation and different walking animations work. 
Problems i found:
- The walking and running animations are not matching the speed of the character
- When the soldier dies and the player leaves the tag area for the soldier it doesn´t despawn. Instead it stays in the dead position and travels back to its original location
- The body of the soldier should despawn when the death animation finishes, not always at the 3 second mark. That way we can avoid the unanimated phase at the end

PR #321, opened 2025-07-30
Task/archer-final-version-tho-780

DHazardFenrir:
New animations and behaviour for the archer. Go to hound of ulster and test the branch with the same name as this on level 1. 

Note: My branch cannot be updated to the new engine release since my desktop computer freezes and gives me a blank screen. If somebody could merge this and test it to see if nothing is broken I would appreciate it very much. 

Note 2: On the bridge, before going to mainland, i tried to do a "machine gun" effect with those archers shooting more arrows, right now  it overwrites the projectiles, that's why it gives a somewhat curious visual effect.


caspi99 (changes requested):
Without looking at the archer, in the code you can see how many changes that have been made during the last month are deleted, like the emissive, the icon or the tileShadingCompute Shader

IDarkgenesis (changes requested):
Tried for a long time to fix and merge but its too difficult, probably better to create a new branch from alpha_2 and port your changes.

PR #324, opened 2025-08-04
THO-781 First Banshee iteration

IDarkgenesis:
This PR add the first iteration for the bansee, it changes the state machine and animations, it also adds some VFX for the scream preparation and attack.

Current Animations: Idle, Search, Teleport, Scream, Hit and Death.

To test: Pull the THO-781-Banshee-final-version from Hound of Ulster repo and delete library, then load level 2 which is where Banshees are located. Scream attack can be cancelled when hit during its warning time, whens its full screaming it can't be cancelled.

banshe_fullAttack_v2


caspi99 (approved):
Good Job ♿

PR #330, opened 2025-08-08
THO-879 Bones in moving uv shaders

IDarkgenesis:
Moving UV shaders now are compatible with meshes using animations, it also adds the option to use shader scripts after the particles and billboards, options is called Post Effects.

movingUV_Anim


PolCurto (approved):
👌

srLlobet (approved):
Looking good!!

PR #331, opened 2025-08-09
Player Healing VFX

PolCurto:
IMPORTANT: This PR is built upon the [New iteration of player final version](https://github.com/TheCenturiaGames/Sobrassada_Engine/pull/328) one, so test that one first.

Added all the custom shaders needed for the heal vfx and its enabling / disabling in the mc script. Their parameters are already as they should be.

TO TEST: go to hound of ulster repo, heal-vfx branch, and pick a mushroom and heal to see the effect. I placed a lot of mushrooms in tutorial and lvl 1 to test this.


MarcLopezS (approved):
Works smoothly! Good Job! :)

PR #332, opened 2025-08-11
PR for changeling / pooka V1

LukasGanter:
Changeling / Pooka implementation. 
There are four different behaviour versions:
- Single dash attack when in range, bites when burried and player is directly above (Version index 1)
- Stays hidden, creeps up to the player when he is looking in another direction, bites when burried and player is directly above (Version index 2)
- Four dash attacks around the player to restrict player movement, bites when burried and player is directly above (Version index 3)
- A combination of the three previous states. Switches between the behavious randomly (Version index 0)

For testing: In the game repo there is a branch with the same name as this one with a changeling test scene. Multiple enemies in different versions are placed there. There is also a current pre release 

Missing features include parameter tweaking, vfx and sound


LukasGanter:
> Looks good! The only thing that seems off, and I don't know if it's intended, is that hitting the buried pooka after it attacks makes it follow me while buried without an animation:
> Grabacion.de.pantalla.2025-08-13.113003.mp4

That is not intended, i´ll try to figure out why he misbehaves

PolCurto (approved):
I still have that issue when attacking while buried sometimes, but it can be fixed in the future and the rest seems really solid, good job!

IDarkgenesis (approved):
Good job, have a look at the feedback where I wrote the issues i found testing the Pooka.
👍

LukasGanter:
> Good job, have a look at the feedback where I wrote the issues i found testing the Pooka. 👍

I will add them in the V2 version

PR #333, opened 2025-08-13
THO-892 Animation scales

IDarkgenesis:
This adds the scaling to the transform for animations which were not used, needed for VFX mesh animations. It also fixes an error that happend during merge that broke the transparent moving uv shader.

If you want to test with a mesh and animation that contains scaling: 
[shoutstart.zip](https://github.com/user-attachments/files/21761145/shoutstart.zip)

movingUV_Anim_scale


PolCurto (approved):
🙏

MarcLopezS (approved):
Works perfectly. Amazing! Good job! 🥇 🔥

PR #334, opened 2025-08-16
Task/New Iteration of player

PolCurto:
This PR adds some new features to the player:
- Player invulnerable while using the ultimate.
- Healing animation cancel on hit & the actual heal is done when the animation explodes, not just at the beginning.
- Attack when key release, so you can charge the attack without doing an attack first.
- Improved VFX of the riastrad transformation.

Also, UI Shaders have been created:
- Life bar shader, which turns from green to red when life gets below half.
- Riastrad bar shader, which has a little wave effect.
- Cooldown shaders, which also have a wave effect, and show the remaining cooldown of the dash and the ultimate.

This wave effect the shaders have can be tweaked in the ShaderScript inspector. It can also be removed if people don't like it, I just did it to test how it looks like. Also, if more bars are added to the UI, it's very probable they can be reused, maybe with little adjustments.

TO TEST: Go the the riastrad-adjustments branch in the game repo and play in the tutorial to see if the changes work. The other scenes don't have the UI updated, so it won't work there. It will be added everywhere when it's merged.
To test the life and riastrad, in the end of the tutorial there is an archer which shoots at you, so you can see the life drop and the riastrad go up. Additionaly, for testing purposes you can hit F10 to fill a little the riastrad bar and fully heal.

### DIFF
--- a/SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/CuChulainn.cpp
+++ b/SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/CuChulainn.cpp
@@ -23,6 +23,7 @@
 #include "SceneModule.h"
 #include "ScriptComponent.h"
 #include "ShaderScriptComponent.h"
+#include "Standalone/AnimController.h"
 #include "Standalone/AnimationComponent.h"
 #include "Standalone/Audio/AudioSourceComponent.h"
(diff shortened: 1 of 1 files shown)


IDarkgenesis (approved):
Tested the mentioned changes and looks good!

MarcLopezS (approved):
Works as expected and very smoothly. Good Job! 🔥

PR #338, opened 2025-08-20
Task/THO-809 video file support

caspi99:
Video support:
Testing: You need to add a video component, you can play and stop. For putting a video you need to create a folder in assets called Videos and you can add any video.
The video does not have audio, we will use wwise system.
Recomendation: Use the video in a empty scene or in a place you dont have too much gameobjects rendering, video is demanding and we cant do threads for opengl

I used the video Carla passed to me, if you want to try

https://github.com/user-attachments/assets/b9c31403-3619-4dd0-94d2-acf423e66efa


SobrassadaEngine/Scene/Components/Standalone/VideoComponent.cpp
```diff
@@ -0,0 +1,309 @@
...
+    }
+}
+
+void VideoComponent::ParentUpdated()
+{
+}
+
+bool VideoComponent::InitVideo()
+{
+    const std::string path = VIDEOS_ASSETS_PATH + std::string(videoName) + VIDEOS_EXTENSION;
```
LukasGanter, line 178: This path works only if your opened project is in a specific location relative to the engine. It should be based on the project path, see other path implementations to check how its done

LukasGanter (dismissed):
Video works, but only with restrictions, mainly due to what is mentioned in the comment. 
Additional things that should be included are: 
- Proper drag and drop import for video files
- Saving of the video file in the library, as we should not package the asset folder into releases

PR #342, opened 2025-08-22
Magic barrier THO-898

LukasGanter:
Implementation for the magic barrier.
Uses the tag system to activate and deactivate depending on how many enemies are left in the fighting area. 

Known issues: The added collision is not yet blocking the player from moving forward. This will be implemented by the mc team

For testing: Open the branch with the same name in the game repo and add the Barrier prefab to the level. Update the tag name input in the script to the one of the island you want to shield. After no enemies are left in the area the barrier will deactivate itself


PolCurto (approved):
Works as expected, good job!

PR #344, opened 2025-08-23
Player hurt screen effect

PolCurto:
First version of the screen effect when getting hurt. When hit, a strong red effect appears for a brief moment in the screen. Also, when you have 1 or 2 hp the effect is constantly present, lighter when 2 hp and heavier when 1.
Also created a new pass for the UI custom shader scripts, which are rendered after the normal UI elements and sorted by the canvas layer.

TO TEST: Go the the riastrad-adjustments branch, and walk towards the archer to get hurt and see if the effect works fine. It should also disappear if you get healed.


DHazardFenrir (approved):
Works as expected 👍

LukasGanter (approved):
Works great, nice job!

PR #345, opened 2025-08-23
New Soldier behaviour and trail

PabioGF:
Now the soldiers wait when there are a certain number of enemies already attacking the player. In the future they will have a cheering animation, now they are just in idle. The number of soldiers before they start to wait can be changed in the inspector, just like the distance when they stop to wait.

I have also added the trail in the sword when attacking and a function to search by the name in the childrens of an object that you can use in the future if you need it.

For test:
Go to the Soldier final version branch in the game repo and go to level 1, I have putted the player in the last part of the level that is were more soldiers are.

### DIFF
--- a/SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/Soldier.cpp
+++ b/SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/Soldier.cpp
@@ -351,7 +351,7 @@ void Soldier::Attack(float deltaTime)
 
             if (animComponent && !animComponent->IsFinished() && thrustAdvance)
             {
-                agentAI->PauseMovement();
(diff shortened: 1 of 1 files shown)


LukasGanter (changes requested):
The basic functionality works well, especially the part when the soldiers stand back, good job!
Some small changes i suggest: 
- The animation movement speed does not fit the movement speed on the ground. Try to reduce the animation speed in the statemachine until it fits
- It happens quite frequently that soldiers follow you while turned to a different direction. Make sure to always call the LookAtMovement function properly

DHazardFenrir (approved):
Apart from the changes lukas requested, its a good job! The animations looks a lot better.

LukasGanter (approved):
Works now, good job

PR #348, opened 2025-08-25
Pooka improvements THO-782

LukasGanter:
Some small improvements to the pookas animations, states and collision behaviours.
Also includes now a last attack functionality that, when there is only a certain number of enemies in the area left, causes all pookas to charge the player and explode near him. This solves the issue that the player has to search remaining pookas before the barrier drops. For now, the explode animation and effect is missing. So the pooka only dies.
Sound is added as well, for testing use the soundbank  provided under the link in discord.

Known issues: 
- The multi dash behaves wrong sometimes (Wrong orientation, first dash in the wrong direction etc. This will be fixed later

To test, open the same names branch in the game repo. Once one pooka is killed all others will do the last charge attack.


PolCurto (approved):
Works as expected. Audio also works nice 👍

MarcLopezS:
Remember to change branch target before merging. Do not confuse in merging main branch :)

LukasGanter:
Entered wrong target branch, will open and merge a new PR

PR #352, opened 2025-08-26
THO-783 water spouts trap final version 2

caspi99:
Updated spouts with some of the feedback.
Testing: Go to branch SpoutsV2, if you walk you will find the spout. Now there's an option of having a water rune in the floor (not necessary to use) but we could use it to have spouts in places there is no water.
We still missing the audio
Sometimes the explosion do weird things, we should do a priority render for transparent objects


SirKaza (approved):
Looks cool, not sure if is expected that physics world and others debug features cannot be seen now. Because of this I'm not sure if damage collider is long enough (sometimes you are not hit if you collision late).
Apart from that good job.

PR #356, opened 2025-08-31
[URGENT TEST] New Mirage Boss Mechanic

srLlobet:
ezgif com-video-to-gif-converter(4)

Allright, here is the new big boss mechanic, the mirage.

Here is how it works: 
There are 3 sequences. which are represented by 3 empty gameobjects. These are the 3 mirage sequences that will trigger when the hp of the boss goes under a defined value. First sequence is at 50 hp, second at 30 and last at 10.

Now, sequences can have a lot of mirages, you can add a mirage prefab to a sequence by making it a child of that sequence.

Mirages have a parameter called weightOrder. This defines the order in which the mirages trigger. For example:

Sequence 1 has 3 mirages weight order 1, 2 mirages weight order 2 and 5 mirages weight order 3:
- When boss hp falls below or is at 50, sequence 1 will trigger. First, the 3 mirages will trigger at the same time. Then, the next 2 and then the next 5 in that order. This means that making a spiral of mirages means that every mirage needs to have a higher weightOrder than the last.

CONSIDERATIONS: 

1- MIRAGE HAS NO VFX currenly
2- Mirages get culled if ofscreeen. This will need to be fixed later. (an idea was to zoom out the camera more for boss) This leads to some weird timings sometimes.
3- There are 3 sequences, please do not add more and leave the names as they are.
4- The big squares are just a vfx representation. They have no actual use, and will not be in the final version at all.

HOW TO TEST:

1. go to branch miragemergetest in sobrassada and MirageImprovementsTest in Hound of ulster
2. To add a mirage, go to desired sequence and add a new mirage prefab as a child to the sequence.
3. Tweak parameters of mirage script to taste
4. click play, fight the boss and enjoy!

I am anticipating a lot of problems because it's quite finnicky, but i will be online all day to solve them and we'll try to merge this slowly tonight with marc.


MarcLopezS (approved):
It works perfect! Good job!

PR #359, opened 2025-09-04
Constant vpo for magical objects

LukasGanter:
Adds a checkbox to the material wind parameters which removes all position or uv based variables from the vpo calculation. 
This results in the possibility to create objects which just gently sway around without deforming, planned for usage in magic environment rocks.

For testing, just apply the wind to any mesh and check the use constant movement checkbox. The mesh should not deform anymore, only move, while all other controls stay functional. 
Also test that this change does not break the normal wind shader when the constant movement is disabled.


MarcLopezS (approved):
Works perfectly! Previous wind still works. Good job! 🎐

PR #361, opened 2025-09-05
Proyectile first version

MarcLopezS:
Now there is a first version of the proyectile. Is not a cube anymore! In order to see the spear in the game, I made the spear mesh a little bit bigger. If not done like this way, it would be hard to see ingame.

Also, when player throws spear, it's weapon mesh is disabled to make it more sense. Once the proyectile disappears, mesh weapon is activated again.

Trail position has been modified in order to not look like a rocket. 🚀 

How to test:
- go to "playerfix" game repository branch.
- try to throw spear ingame.

(this is a first version. If you see something to improve, all feedback is welcome 😄 )

### DIFF
--- a/SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/CuChulainn.cpp
+++ b/SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/CuChulainn.cpp
@@ -964,8 +964,7 @@ void CuChulainn::UpdateTimers(float deltaTime)
         {
             weapon->SetEnabled(true);
             resetWeapon = false;
-            GameObject* spearCharacter =
-                AppEngine->GetSceneModule()->GetScene()->GetGameObjectByName("WP_Spear_Cu_Chu");
+            GameObject* spearCharacter = AppEngine->GetSceneModule()->GetScene()->GetGameObjectByName(spearNameGOMesh);
             spearCharacter->GetComponent<MeshComponent*>()->SetEnabled(true);
         }
         throwTimer = 0.0f;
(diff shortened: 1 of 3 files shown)


SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/CuChulainn.cpp
```diff
@@ -964,6 +964,8 @@ void CuChulainn::UpdateTimers(float deltaTime)
         {
             weapon->SetEnabled(true);
             resetWeapon = false;
+            GameObject* spearCharacter = AppEngine->GetSceneModule()->GetScene()->GetGameObjectByName(spearNameMesh);
+            spearCharacter->GetComponent<MeshComponent*>()->SetEnabled(true);
```
PolCurto, line 968: Probably better to store this in the Init(), so you don't have to get the gameObject by name each time which can be expensive

SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/CuChulainn.cpp
```diff
@@ -1118,10 +1120,13 @@ void CuChulainn::ThrowSpear()
     {
         weapon->SetEnabled(false);
         resetWeapon = true;
+        GameObject* spearCharacter = AppEngine->GetSceneModule()->GetScene()->GetGameObjectByName(spearNameMesh);
+        spearCharacter->GetComponent<MeshComponent*>()->SetEnabled(false);
```
PolCurto, line 1124: same here

PolCurto (approved):
Works as expected, good job! Take a look at the comment before merging. Also, maybe I would even make the spear bigger, because with the trail I can barely see it

PR #364, opened 2025-09-05
On-Hit VFX V1

PolCurto:
First iteration of the VFX on enemy hit

New features:
- Added a spritesheet effect that appears when an enemy gets hit in the opposite direction of the player. 
- New shader that turns enemies to a plain color for a brief period of time when hit, by default white. You can change the color in the ShaderScript inspector
- TimeStop when enemies are hit or die, being longer when dead. You can tweak the duration in the enemies' script inspector

All changes have been made to the Character class, so it is applied to all enemies at once.

TO TEST: Go to the om-hit-vfx game branch. In the Level 1, there is one enemy of each at the start to test it. You can go to the boss level to see the boss


SirKaza (approved):
Nice changes.
Only problem that I see is the boss's ponytail is not affected by the VFX because is another mesh but for now it works

PR #370, opened 2025-09-07
THO-861 Crash between scenes

IDarkgenesis:
The crashses should now be fixed, hopefully, tested in level2 before boss scene at full performance and at 12 fps, and did not crash.


caspi99 (approved):
It didnt crashed to me

SirKaza (changes requested):
image

It crashed for me from level 2 to boss level, hound of ulster branch alpha2

SirKaza (approved):
I tested now several times and didn't crash. 👍

PR #376, opened 2025-09-08
Archer Revision-THO-924 

DHazardFenrir:
To test this PR to check the new behavior of archers is branch v3/task/ArcherRevision. In Hound of Ulster is the branch v3/task/Archer-final-version-THO-780. The archers should chase, shoot, and have textures and sound, the walls also block projectiles.


SirKaza (approved):
The trail of the arrow at the start does some weird curves sometimes and an archer keep stuck running at the same position.
Apart from that, fantastic job!
I like the different behaviors and the multiple projectiles 👍

PR #379, opened 2025-09-09
bug/Player projectile

PolCurto:
Fixed the changes that broke the player projectile.

TO TEST: Go to alpha 2 and test that the projectile works. Also test that archers still work fine.


MarcLopezS (approved):
For me it worked. Good job! 👍

PR #383, opened 2025-09-09
fix arrow crash

PolCurto:
I couldn't test it because to me in alpha 2 archers don't shoot and teleport around, but this should fix the crash


LukasGanter (approved):
Fixes the crash, good work

PR #385, opened 2025-09-09
Kinematic physics transform wrong calculation

IDarkgenesis:
This PR should fix the rotation of the physics colliders, for example the spear projectile or projectiles when shot, now they should follow the game object rotaion


PolCurto (approved):
🦯

PR #386, opened 2025-09-12
Scene loading times reduced THO-925

LukasGanter:
This pr reduces scene loading times by caching metadata file paths.
A cache invalidation is also added to incorporate newly imported files into the cache as well.
Further time decreases might be possible but are not necessary for now as we will load async while playing story videos during scene switches.

For testing, just open the engine and load other scenes and see if you can feel an improvement. 
Also, import new gltfs and see if they can be added to the scenes without an engine restart.


PolCurto (approved):
Loading time are way faster, good job! Also, I've been able to import a new gltf and load it without issues

PR #387, opened 2025-09-14
Player bug fixing (floor clipping and attack vfx on death)

PolCurto:
Fixed some errors in the CharacterController that led to floor clipping in some circumstances (modifying the position in y in some situations it should not). Also now if you die while attacking the attack vfx should be disabled.
Also as some teachers suggested now if there is an enemy close in front of the player, it won't move forward when attacking.

TO TEST: Mainly test that the floor clipping issues are solved, as it is one of the most game-breaking issues right now. Dash everywhere and try to break it. If it happens tell me where and I'll take a look. Also, enable the navmesh debug view, because in some places this navigation issues are because of the navmesh, not the player.
For the other things, just test if it works as it should.


DHazardFenrir (approved):
Works as expected 🧿

PR #389, opened 2025-09-17
Debug library caching THO-1000

LukasGanter:
This PR fixes a crash when using the debug mode, it might need a full engine and project library rebuild to work properly.


caspi99 (approved):
Works! Good Job!

PR #390, opened 2025-09-18
THO-1001 Raycast to physics world

IDarkgenesis:
This PR adds the function to raycast into the physics world, it return nullptr if nothing was hit, if there is a result, it returns the BulletUserPointer of the object.

I have added a call to the raycast at the CheckDashObstacles at the CharacterControllerComponent , to test it you can go to Tutoral and dash against object that contain a collider component, such as mushrooms or the soldiers. Once this PR gets accepted I will remove the code.

### DIFF
--- a/SobrassadaEngine/.vscode/settings.json
+++ b/SobrassadaEngine/.vscode/settings.json
@@ -1,6 +1,0 @@
-{
-    "files.associations": {
-        "xiosbase": "cpp",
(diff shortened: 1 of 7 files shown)


MarcLopezS (approved):
It works perfectly! Just what I needed. Good job! 🔥

caspi99 (approved):
🩻

PR #391, opened 2025-09-18
FXAA Anti-Aliasing THO-811

PolCurto:
Implemented the FXAA anti-aliasing algorithm. It takes place in the RenderPass after all the other passes, but its previous to the UI.

In View -> Editor Settings -> FXAA dropdown, you can find some parameters to tweak:
- Enable / disable FXAA
- Show borders (show the edges the algorithm is smoothing, horizontal in green and vertical in red)
- Global and local thresholds, the smaller they are the more smoothed edges will be. By default are the 'optimal' values that show the slides

There is also the option to enable / disable it in the Game Debug menu.

TO TEST: check if it works and if there is a performance impact (it shouldn't). For my experience, in the engine is noticeable with further objects, but in the game release the difference is minimal.


IDarkgenesis (approved):
Works both in Engine and Game, good job! 👍

DHazardFenrir (approved):
I had no problem testing, works as expected! 🪅(it didn't have an impact on performance on my side as well).

PR #392, opened 2025-09-20
THO-1002 transparency order

caspi99:
This PR scares me. It solves problem with transparent objects covering other transparent objects. We solve it not writting on the z-buffer (we should not do it for transparency). Some meshes can have strange things now, for example, the boss phere in the boss is really transparent now, but we can solve it putting the sphere opaque, but it solves the general problem. 
Also solves the problem with some black pixels in transparent objects, not premultiplying the meshes.

image
Things like this image wont happen again

image
We will see it like that now


PolCurto (approved):
🩻

PR #393, opened 2025-09-21
Boss Spouts + Mirage VFX THO-889, THO-885

srLlobet:
Allright! Double Pr (sorry Lukas)

TO TEST: Use hound of ulster branch: SpoutsRe as it has the new Ferdiad statemachine, boss spouts and new mirage vfx

Water spouts are here! They start appearing in phase 2, and can be mixed with other attacks. They spin around the arena and all parameters are customizable.

Mirage VFX is also here! In the branch, there are only four mirages appearing in phase 1, so you can test it quickly. The colors can be changed however you want.


PolCurto (approved):
Works fine, good job! Just one thing, shouldn't the mirage vfx arrow go way faster?

srLlobet:
I'm only adding the patterns on the final version of the arena in the hound of ulter repo, as for the other changes, mirage vfx speed is also hound of ulster, (animation speed parameter, 0.5 looks normal speed, and the more, the faster). Removed rune on boss and bubbles on the boss spouts too.

Height should be fine to change by scaling the spout, but I'd recommend waiting until the final arena.

LukasGanter (approved):
The changes work, please implement the game repo related features as soon as possible as well for testing

PR #400, opened 2025-09-24
Exponential height fog THO-725 & Fix FXAA debug render THO-1004

PolCurto:
Implemented exponential height fog. As with FXAA, there are parameters you can tweak in View -> Editor Settings -> Height Fog
image

You can:
- Enable / Disable
- Fog follow camera height. If this is checked, the fog will always be at the height specified by the fog start heigh depending on the camera, meaning that if the cameragoes up, the fog goes up and viceversa. If unchecked, the fog is always at the same height, inependently of the camera position.
- The global density of the fog
- The heigh fallof (the smoother it goes from 0 to 1 amount of fog)
- Max fog (minus than 1 make it not opaque)
- The fog start height, to make it higher / lower
- The fog color

All this parameters are saved into the scene when the scene is saved, so each level can have its fog customized.

IMPORTANT: the fog is rendered previous to transparent objects, because as transparent objects don't write to the depth buffer, if the fog goes after it is always drawn on top of them, and we don't want that because then it covers things like the attacks vfx even if the fog should be in the background. Because of this, there are some transparent objects under the fog that can be seen, so they should probably be removed.

TO TEST: Play the game and tweak fog parameters to see if it works fine. There is a branch in the game repo called height-fog with already fog customized for each level.

Also fixed the FXAA Anti-aliasing blocking the debug view. The problem was that it was writing to the depth buffer, and as it is a fullscreen quad it blocked everything after that. Just test if it does indeed work now.

### DIFF
--- a/SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/Changeling.cpp
+++ b/SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/Changeling.cpp
@@ -143,7 +143,7 @@ void Changeling::OnPlayerEnterLocation()
 void Changeling::PlayHighlightSequence()
 {
     // Don´t play the highlight if the pooka is already doing something else
-    if (currentState == ChangelingStates::IDLE_BURIED)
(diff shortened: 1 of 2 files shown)


DHazardFenrir (approved):
👀👍

PR #401, opened 2025-09-25
THO-949 shadow render fix

caspi99:
For me shadows are working good now. 
Also I added a little more smooth in the PCF and I added a shadow Strength variable (PCF also reduces the shadow intensity).

image

For testing: Just play the game. You can touch the directional to see better the shadows (In level 1, there is more than 1 directional, the engine only support one at the same time, you can try other levels or gl searching)


DHazardFenrir (approved):
Seems like (another) render pipeline bug the one I encountered. Since now after turning up the laptop again the error isn't replicable, and Andreu told me the error popped up but when there were no shadows, this seems like an error that appears under random conditions.

SirKaza (approved):
Great work, everything works well!
We should do shadow strength 100 and swap to a horror game 🕶️

PR #402, opened 2025-09-26
THO-729 Volumetrics

IDarkgenesis:
This adds volumetric fog that can be rendered using spotlights, right now there is a limit to 15 spotlights that will calculate volumetrics per scene. In order to the volumetrics to be rendered there needs to be a VolumetricAreaComponent inside the scene and it must intersect with the spotlight frustum, AABB and Sphere shapes are supported. 

Right now parameters are set by scene, but it can be moved per VolumetricAreaComponent to have more customization of the volumetrics, the anisotropy parameter takes a huge rol into how the light is scattered.

In order for this to work shadow maps for spotlights are created (15 max, can be changed), I have not put the shadow code for spotlights but it could be added later, the work of calculation is already done.

To acces the config for fog -> View->Scene->Fog Config. There is a noise function in the shader, but a texture can be used.

To test: Load a scene place some spotlights behind some meshes and put a volumetric are in front to see the effects.

image

image

image


SirKaza (approved):
Well done!
Only thing that seems off is that sometimes I have to resize engine window to enable god rays (strange bug). Apart from that, everything seems fine.

PR #403, opened 2025-09-29
Finalized version for the destructibles

LukasGanter:
This PR changes the behaviour of the destroyed mesh (Appears with a delay, stays visible forever).
Sound is also added depending on the type of destructible.

For testing, go to the same branch in the game repository. An example vase is located right in front of the player in level 2.

### DIFF
--- a/SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/Destructible.cpp
+++ b/SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/Destructible.cpp
@@ -8,10 +8,10 @@
 #include "Globals.h"
 #include "ParticleSystemComponent.h"
 #include "ResourceStateMachine.h"
-#include "Wwise_IDs.h"
 #include "Standalone/AIAgentComponent.h"
-#include "Standalone/MeshComponent.h"
 #include "Standalone/Audio/AudioSourceComponent.h"
+#include "Standalone/MeshComponent.h"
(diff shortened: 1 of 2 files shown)


DHazardFenrir (changes requested):
Everything works perfectly, I didn't have any problem with the destructible. BUT I didn't hear any sound when breaking it. In the scene in the game branch there is an audio source but not audio selected, but even if I selected audio just for testing, it didn't play at all. 

If the audio source I mark the checkbox 'play on start' then it plays. Also, even if I can see the new audio files are in fact in Wwise.h, those aren't displayed in the engine (and the task seems duplicated at least in the Github Desktop GUI). 
image
image
(I deleted the library to see if that fixed the sound but no). I am on the respective engine and game branch.

LukasGanter:
> Everything works perfectly, I didn't have any problem with the destructible. BUT I didn't hear any sound when breaking it. In the scene in the game branch there is an audio source but not audio selected, but even if I selected audio just for testing, it didn't play at all.
> 
> If the audio source I mark the checkbox 'play on start' then it plays. Also, even if I can see the new audio files are in fact in Wwise.h, those aren't displayed in the engine (and the task seems duplicated at least in the Github Desktop GUI). image image (I deleted the library to see if that fixed the sound but no). I am on the respective engine and game branch.

Are you using the most recent version of the soundbank? I updated it yesterday in the evening

DHazardFenrir (approved):

PR #405, opened 2025-09-29
Button-image-rendering-THO-1007

jmartiolivart:
This PR unlinks button and image to avoid the bug in the menus, I also added sounds to all the menus throught code and changed some order canvas logic.

To test this go to The hound of Ulster and Sound-Menus branch (it should start at MenuTutorial).

**In hierarchy/inspector create a button (it should be possible to create it without a image attached)**

Some of the things added:
*MainMenu Level*
-  All button sounds in tutorial menu
- Fix image bug

*Tutorial Level*
- Added sound buttons on Pause and on Gameover
- Fixed images bug in this menus

(i added 5 damage to the first soldier to be easy to die)


PolCurto (approved):
Menus and sounds work as expected, good job! The sound when selecting something sounds too late, but that can be easily adjusted later in wwise

SirKaza (approved):
Nice changes!
The main menu only supports directional arrows in keyboard and not WASD while other menus does support them. I think it's just missing in the main menu.
Also, take a look at the comments below.

PR #408, opened 2025-09-30
Ultimate VFX Animation

MarcLopezS:
Ultimate attack has been fixed. Now it is not interrupted while player finishes its animation. Also, sounds and some meshes from ultimate attack have been updated in order to not look weird.

While ultimate attack, input controls are disabled in order to not expect weird behaviour ingame.

How to test:
- Go to Ultimate animation fix game branch.
- Use ultimate and study its behaviour.

NOTE: some VFXs from ultimate have been tweaked in order to work, but a render bug in engine is still present. Once that bug is fixed, some VFX will again be updated.


PolCurto (approved):
Works fine, good job! I think it lasts too long but apart from this works as expected

LukasGanter (approved):
Works good, nice job. As Pol said the duration should be a bit shorter, also the vfx are floating above the ground. This can be fixed aside from this pr though

PR #409, opened 2025-09-30
THO-915 Particle system culling

IDarkgenesis:
Particle system AABB now is computed taking into account the maximum possible space covered by its different emitters, it also fixes Spritesheets not working as intended.

To test: Load any level and check PS are not broken and also for example in lvl 1, fires should not pop in and out of existence that abruptly.

spriteWorking


PabioGF (approved):
Works perfectly, good job!

PR #410, opened 2025-10-03
New player iteration & HUD

PolCurto:
This PR adds a lot of new player vfx, hud changes and animations:

VFX:
- VFX when charging attack and when releasing
- New VFX For attack 1 & 2
- New VFX when transforming to riastrad
- New screen effects when getting hurt
- New dust VFX when dash
- New VFX for ranged attack (WIP, will be updated in the future)

HUD changes: 
- The left red orb is now filled only when riastrad bar is full. When riastrad is triggered, a burning vfx appears in the orb, and when it finishes its emptied. The key to trigger riastrad appears when its full, so the player always has in mind how to use it.
- The right circle now has the amount of mushrooms, as empty or filled orbs. Whn a mushroom is picked up, an effect plays and an orb is filled. When used the same, and effect is played and the orb disappears.

Player animations:
- Idle during riastrad is different now
- Running during riastrad is different now
- Walking during cursed state is different now
- Pick up mushroom
- Ranged attack now is split in 2 animations, which avoids the weird aiming in wrong poses. Yes, the animation still does not aim to the proper direction.

Also different fixes and adjustments for the player, and, now in the state machine state editor, you can input text to search for the clip name

TO TEST: Go to the new-player-vfx branch of the game repo and check the listed things work properly. Changes are only applied in the tutorial level. There were a lot of additions so maybe I forgot something


MarcLopezS (approved):
Very cool and beautiful! Good job! 🔥

PR #414, opened 2025-10-04
Task/Soldier Finalization tho 958

PabioGF:
The final version of the soldier.

To test it go to the branch with the same name in the game repo

Now they are asigned randomly the helmets and the material, that can be brown/red or blue.
Added also VFX for the attacks and the different sounds like damaged, death, and detect player.

test that all goes smooth


PolCurto (approved):
Changes work fine, good job! However, I would personally make the soldier attack from further away, and have a bigger hitbox that is justified with bigger vfx, because right now they have to get so close to you that they never attack except if you let them

DHazardFenrir (approved):
Works good, apart from what pol said that about the soldiers, I would add to attack them faster or interrupt the combo if you aren't near, because if not, you can just dodge and they would be easier to kill.

PR #415, opened 2025-10-05
Tutorial input device UI  THO-916

LukasGanter:
This PR adds the functionality to change the tutorial UIs depending on the selected input devices. 
Supported are keyboard, dualshock and xbox, with the xbox ui showing for all controllers which aren´t dualshocks.
Also adds the option to show ui only once, and for it to disappear after x seconds.
I don´t have an xbox controller for testing, so at least one tester should test with that.

For testing, go to the same branch in the game repo. The movement tutorial works like previously, only showing the correct controls for the input device. The dash tutorial is triggered only once and disappears after 5 seconds

Known issues: 
There are ui elements in the main menu and health bar which are not showing the proper controls. These will be tackled with a later pr / change in the scene

### DIFF
--- a/SobrassadaEngine/Modules/InputModule.cpp
+++ b/SobrassadaEngine/Modules/InputModule.cpp
@@ -278,6 +278,15 @@ void InputModule::OnControllerDisconnected()
     }
 }
 
+bool InputModule::IsPlaystationControllerConnected() const
+{
+    if (controllers[0] == nullptr) return false;
+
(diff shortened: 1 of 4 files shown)


srLlobet (approved):
It works fine! It does detect the controller, and the dash message leaves after a while. Good job! (Not tested with xbox controller but dualshock and keyboard work fine)

PolCurto (approved):
I tested with Xbox controller and works fine as well

PR #417, opened 2025-10-05
Destructibles V3 THO-1012

LukasGanter:
This PR changes the base class of the destructible script from a character to a normal script to reduce complexity and increase performance. 

For testing, go to the same branch in the game repo and add the All_Tributes_Destructible prefab to a scene. The destructibles should work like normal while also playing sounds now


DHazardFenrir (approved):
I had to rebuild the library, but once it was done, everything worked just fine and as expected.

srLlobet (approved):
Yep! Working nicely! Good job

PR #420, opened 2025-10-06
New pooka vfx

PolCurto:
This PR adds new changes to the pooka as well as VFX:
- The final attack is replaced by the pooka just attacking all the time with his normal dash.
- After burying, if the player was in sight, the pooka moves underground to the last player location
- VFX above the dash trail, on dash start and when dash finishes
- VFX when biting the player
- VFX on damage
- VFx when burying down

TO TEST: Go to lg_pooka-finalization branch of the game repo, there is a Pooka in the beginning of the tutorial, and test that it works fine and new VFX are properly seen.

Also now spritesheets can have variations


MarcLopezS (approved):
Nice work! 
When testing, if player pass through the pooka dash trail, the animation displayed for player is idle instead of running. Also it happened a weird noise in the SFX (only happened once). Since this PR is only for the pooka VFX, I approve the PR, but think about this to fix it in a future PR. Appart from that, good job! 😄

DHazardFenrir (approved):
I had no problem with the pooka vfx, but yes, when stepping on the trail the player get stuck on idle state/invulnerable, but the VFX for the pooka are nice and work properly.

PolCurto:
I'm pretty sure that's because the game branch is not updated with the new player animations

PR #426, opened 2025-10-07
THO-1015 Engine UI flicker

IDarkgenesis:
This should fix the engine flicker when resizing the imgui ui.


PolCurto (approved):
Not flickering anymore 🦈

caspi99 (approved):
🛬

PR #427, opened 2025-10-07
Adaptive music THO-1014

LukasGanter:
This PR adds scripts to dynamically control which music track is played in which level and area. 
Once the player enters a dedicated collider the music will change smoothly. 

For testing. open the same branch in the game repo. When spawning, no music is playing. If you continue to walk through the level music will start, and then switch to different tracks once at the end of the fireball area and once at the tree of life behind the bridge. The associated colliders disappear once activated. They are all enabled again once the player respawns (Can no be tested directly as the ui is broken)

Known issues: If there are compile errors in the soldier script just comment out the lines for testing, this problem will disappear with the alpha 3 merge


IDarkgenesis (approved):
Sounds change as expected, good job! 🔊

PolCurto (approved):
Works as expected 👍

PR #428, opened 2025-10-07
Task/WeHateAssetsFolder-THO-900

SirKaza:
No assets folder needed anymore for playing.
Now audio and videos need to be outside of assets folder. See the following image.
image

One thing to consider, when importing something that creates metadata, it is also created in library folder, but if assets folder is deleted, it will stay only in library and pushes to GitHub will not include these new metadata files.

How to test:
Build library running once, then stop engine, delete assets folders in hound of ulster and engine and running and playing the game.
Also, test moving things and saving the scene, or doing things with metadata.


caspi99 (approved):
Seems all working now, good job!

PR #431, opened 2025-10-08
Additional music triggers THO-1021

LukasGanter:
Adds all necessary script changes for changing the music based on the current screen display.
- Music is starting to play from spawning
- Pause / Death menu has its own music. It changes to it when the menu opens and goes back to the previous music when it closes again
- Riastrad has its own music while activated
- When the barrier disappears the music resets to exploring
- On respawn the music resets to exploring

For testing, go to the same branch in the game repository and test the points mentioned above.

Known issues: For me, the animation of the fireball trap is now extremly fast. I don´t see a connection to my changes, so please tell me if you encounter the same issue

### DIFF
--- a/SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/PauseMenuScript.cpp
+++ b/SobrassadaEngine/Scripting/SobrassadaScripts/SobrassadaScripts/PauseMenuScript.cpp
@@ -81,6 +81,10 @@ void PauseMenuScript::Close()
     isOpen    = false;
     builtOnce = false;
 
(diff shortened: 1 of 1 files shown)


PolCurto (approved):
Works fine, good job! Only thing I found: if you go to riastrad, and pause the game, when unpause the music is the default one even the player is still in riastrad mode

IDarkgenesis (approved):
Music triggers as intended. 👍 

There is a crash when using the player ultimate, highly doubt is anything related to this. And also the fireball anim is too fast, which you commented.
image

